PR #6, opened 2022-08-08
Document NeoVim version requirement

chnn:
Hello! I gave this plugin a try recently, but it errored at start up. The issue is that the plugin uses the `vim.fs` APIs, which are not yet available on the latest nvim release (0.7).

This commit updates the README to make this requirement clear. Feel free to close or modify however you see fit, of course. I'm just opening this as a heads up.

The plugin looks really nice!


epwalsh:
Ah, good catch! Thanks @chnn!

Issue #15, opened 2022-09-22
Allow customizing daily notes path


metalelf0:
### 🚀 The feature, motivation and pitch

In Obsidian, I customized my daily notes path to a subfolder of the main space. I'd like to be able to set it as a custom option in `obsidian.nvim`, so that when I type `:ObsidianToday`, it creates the daily note in that path instead of the root of the obsidian space.

### Alternatives

Right now I'm manually moving the daily note to the specified path, but if I already created it inside Obsidian, I have to manually merge the two files and it's not the ideal workflow.

epwalsh:
Hey @metalelf0, I'd be happy to have this feature. Are you interesting in making a PR?

You'd just have to edit the command function here:
https://github.com/epwalsh/obsidian.nvim/blob/e3d10fe6bf19f99e7ce289a13f0fc720c2cb276f/lua/obsidian/command.lua#L52-L59

And then add a field for this setting in the [Obsidian client class](https://github.com/epwalsh/obsidian.nvim/blob/e3d10fe6bf19f99e7ce289a13f0fc720c2cb276f/lua/obsidian/init.lua#L20-L28) and [config](https://github.com/epwalsh/obsidian.nvim/blob/e3d10fe6bf19f99e7ce289a13f0fc720c2cb276f/lua/obsidian/config.lua#L1-L44).

metalelf0:
Hey @epwalsh ! Sure thing, but my lua-fu is still not that great. I tried hacking a solution here:

https://github.com/metalelf0/obsidian.nvim/commit/dc9956a24ca1c4235f406bcd22bb237f7dc878c9

But it seems something is not working.

Here's my reasoning: I added a client options called `dailies_path`, that should be relative to the main dir. This has a default value of `""` (empty string), so that it can seamlessly be joined to the dir. In the client initialization, I convert it to a `dailiies_dir` variable, so that it can be used without `if`s in the command. Loading the plugin with my changes, however, breaks with or without the option set.  

What am I doing wrong?

Thanks a lot for your work!

epwalsh:
Hey @metalelf0, can you open a draft PR so I can leave suggestions? Also the last few commits I made yesterday touch some of the same code, so you'll want to update your branch first.

epwalsh:
Also I was thinking instead of adding a top-level field like `dailies_path: ""`, we should add a field `daily_notes: { folder: "", ... }` that mimics the exact JSON structure in `~/.obsidian/daily-notes.json`.

[...]

PR #21, opened 2022-09-27
feat: add daily_notes.folder option

metalelf0:
Hey, sorry for the delay, I've been quite busy recently.

Here I added the option to customize the dailies path. Let me know if you think this solution is ok, it should be much cleaner than the previous one. I copied your implementation for the `notes_subdir` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added `daily_notes` configuration options.
+
 ### Fixed
 
 - Fixed a bug with `:ObsidianOpen` ([#19](https://github.com/epwalsh/obsidian.nvim/issues/19)).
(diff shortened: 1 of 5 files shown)


epwalsh (commented):
Hey @metalelf0 this is a great start! Did you see my comment [here](https://github.com/epwalsh/obsidian.nvim/issues/15#issuecomment-1256406320)? What do you think about changing `dailies_path: ""` to `daily_notes: { folder: "", ... }`, i.e. mimic the JSON structure of `.obsidian/daily-notes.json`?

metalelf0:
Hi, I read it but I was unsure whether to follow that approach or to keep the same naming convention of `notes_subdir`. If you think it's more important to mimic the Obsidian JSON configuration, I will change it!

epwalsh:
>  If you think it's more important to mimic the Obsidian JSON configuration, I will change it!

I do think that would be nice because it presents a natural way to add support for more core Obsidian plugins in the future. We could also automatically load that part of the config from the corresponding `.obsidian/*.json` file.

epwalsh (approved):
This looks great! Thanks so much @metalelf0 :)

Issue #24, opened 2022-10-07
Links to notes with spaces in the title are not able to be followed


shakesbeare:
### Summary
Links with space in the note title are not followable by a simple `gf`. Instead, user must highlight the entire contents of the link at which point `gf` works. 

### Steps to reproduce
1) Create a note in vault
2) Link to a note with which has a space in the name (Ex: "From the Vast - Session 1.md") 
3) Place cursor inside the link at any position
4) Use `gf` command
5) Observe error 'Cannot find file "From" in path'
6) Use visual mode to select entire file name
7) Use `gf` command
8) Observe link followed correctly

### Expected Behavior
Link is followed no matter if visual mode technique is used or not. 

### Versions

```
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.8.0/share/nvim"

Run :checkhealth for more info
```

```
1.3.0
```

epwalsh:
Hey @shakesbeare, I'm not sure what the best solution to this would be. I guess we could potentially override `gf` with our own implementation.

epwalsh:
Contributions welcome on this. I looked into a little more and I'm pretty sure the best solution is in fact to override `gf` with a custom function. I think it's reasonable to assume that we can only handle filenames with spaces when they're contained in a reference, e.g. `[[my note.md]]`. So implementing this function shouldn't be too hard. Basically:
1. Get the cursor position and corresponding line in the buffer.
2. Check if cursor is on a reference, in which case parse out the reference and follow it.
3. Otherwise just treat the current "word" under the cursor as a filename and follow it.

shakesbeare:
@epwalsh I figured I'd give it a shot but I had some questions about the exact implementation which is best. 

My thought was to create a command such as `:ObsidianFollowLink` and a companion function, something like `cursor_inside_markdown_link`. That way the user could decide to bind it at their own discretion and choose to use the `gf` keybind if desired and allow it to pass through to standard `gf` if the cursor isn't inside of a markdown link. The idea would be to avoid rewriting vim's existing code for this purpose and allow greater customization for the user. 

[...]

PR #39, opened 2022-10-15
Follow Markdown Link

shakesbeare:
#24 

Everything is working as far as my testing could tell. I'm slightly concerned that I over engineered the solution a bit, though. I wanted to make sure that the links could find the file regardless of whether it was in a sub-directory or not plus account for functionality if the link contained a complete reference to the path of the  note. The solution I landed on was to just scan the entire vault directory looking for a matching note and routing there, unless the note name was ambiguous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,12 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
 - `:ObsidianNew` to create a new note.
   This command has one optional argument: the title of the new note.
 - `:ObsidianSearch` to search for notes in your vault using [ripgrep](https://github.com/BurntSushi/ripgrep) with [fzf.vim](https://github.com/junegunn/fzf.vim) or [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).
-<<<<<<< HEAD
   This command has one optional argument: a search query to start with.
 - `:ObsidianLink` to link an in-line visual selection of text to a note.
   This command has one optional argument: the ID, path, or alias of the note to link to. If not given, the selected text will be used to find the note with a matching ID, path, or alias.
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
-=======
 - `:ObsidianFollowLink` to open a note or create a new note in your vault
->>>>>>> 9720ba9 (Updated README to reflect changes)
 
 ## Setup
 
--- a/README.md
+++ b/README.md
@@ -21,11 +21,15 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
 - `:ObsidianNew` to create a new note.
   This command has one optional argument: the title of the new note.
 - `:ObsidianSearch` to search for notes in your vault using [ripgrep](https://github.com/BurntSushi/ripgrep) with [fzf.vim](https://github.com/junegunn/fzf.vim) or [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).
+<<<<<<< HEAD
   This command has one optional argument: a search query to start with.
 - `:ObsidianLink` to link an in-line visual selection of text to a note.
   This command has one optional argument: the ID, path, or alias of the note to link to. If not given, the selected text will be used to find the note with a matching ID, path, or alias.
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
+=======
+- `:ObsidianFollowLink` to open a note or create a new note in your vault
+>>>>>>> 9720ba9 (Updated README to reflect changes)
 
 ## Setup
 
@@ -195,4 +199,20 @@ require("obsidian").setup({
     return out
   end,
 })
+```
+
+#### Mapping ObsidianFollowLink with passthrough
+```lua
+vim.keymap.set(
+    "n",
+    "gf",
+    function()
+        if require('obsidian').util.cursor_on_markdown_link() then
+            return "<cmd>ObsidianFollowLink<CR>"
+        else
+            return "gf"
+        end
+    end,
+    { noremap = false, expr = true}
+)
 ```
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
   This command has one optional argument: the ID, path, or alias of the note to link to. If not given, the selected text will be used to find the note with a matching ID, path, or alias.
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
-- `:ObsidianFollowLink` to open a note or create a new note in your vault
+- `:ObsidianFollowLink` to follow a note reference under the cursor.
 
 ## Setup
 
--- a/lua/obsidian/command.lua
+++ b/lua/obsidian/command.lua
@@ -174,6 +174,7 @@ command.search = function(client, data)
   end
 end
 
+<<<<<<< HEAD
 command.link_new = function(client, data)
   local _, csrow, cscol, _ = unpack(vim.fn.getpos "'<")
   local _, cerow, cecol, _ = unpack(vim.fn.getpos "'>")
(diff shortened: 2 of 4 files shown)


lua/obsidian/util.lua
```diff
@@ -266,4 +266,22 @@ util.table_length = function(x)
...
 end
 
+-- Determines if cursor is currently inside markdown link
+-- @return integer, integer
+util.cursor_on_markdown_link = function()
+  local current_line = vim.api.nvim_get_current_line()
+  local cur_row, cur_col = unpack(vim.api.nvim_win_get_cursor(0))
+  local current_line_lh = current_line:sub(1, cur_col)
+  
+    print(cur_row, cur_col)
```
epwalsh, line 276:
```suggestion
```

lua/obsidian/util.lua
```diff
@@ -266,4 +266,22 @@ util.table_length = function(x)
...
+  local open = current_line_lh:find("%[%[[^%[]") 
+  local close = current_line:find("%]%]", cur_col)
```
epwalsh, line 278: There's a situation where this results in a false positive. Say the current line has two links:

```
here's a link [[abc]] and here's another link [[xyz]]
```

And the cursor is somewhere in between the two links. So I think you need to do something like this:

```suggestion
  local open = current_line_lh:find("%[%[[^%[%]]*$") 
  local close = current_line:find("^[^%[%]]*%]%]", cur_col)
```

lua/obsidian/command.lua
```diff
@@ -168,13 +168,73 @@ command.search = function(client, data)
...
+        if ok and note:fname() == note_name then
+          notes[count] = entry
+          count = count + 1
+        end
+      end,
+    })
+  end
+
+
+  if table.getn(notes) < 1 then
```
epwalsh, line 217: Since `notes` is an array you can do
```suggestion
  if #notes < 1 then
```

lua/obsidian/command.lua
```diff
@@ -168,13 +168,73 @@ command.search = function(client, data)
...
+  
+  if note_name:match("|[^%]]*") then
+    note_name = note_name:sub(1, note_name:find('|'))
+  end
+
+  local notes = {}
+  local count = 1
+
+  if not note_name:match("/") then
+    scan.scan_dir(vim.fs.normalize(tostring(client.dir)), {
```
epwalsh, line 199: I think it would be a lot more efficient to scan only directories (use `only_dirs = true`) and then check if the given note name is in each directory.

Alternatively, we don't necessarily need to check every directory. We can just check the same directories that are added to `path`: https://github.com/epwalsh/obsidian.nvim/commit/f08be13e846c35efb0afff37b1ed00033374013c#diff-f3def1404e2ac8d33b653d93273b328b44defe45491bf8589b7598809cb6b854

lua/obsidian/command.lua
```diff
@@ -168,13 +168,73 @@ command.search = function(client, data)
...
+  local note_name = current_line:sub(open + 2, close - 1)
+  local path = client.dir
+  
+  if not note_name:match("%.md") then
+    note_name = note_name .. ".md"
+  end
+  
+  if note_name:match("|[^%]]*") then
+    note_name = note_name:sub(1, note_name:find('|'))
+  end
```
epwalsh, line 193: You'd probably want to switch the order of these blocks?

epwalsh (changes requested):
Hey @shakesbeare, this is a great start. I like this approach. Just have some minor comments 👇

lua/obsidian/command.lua
shakesbeare: Does the notes_subdir config allow the user to specify multiple subdirectories? 

It makes sense that they might want to dump all new notes into a single folder when creating them, but they might have multiple top level folders to actually store notes in which they may want to link to.

epwalsh: > It makes sense that they might want to dump all new notes into a single folder when creating them, but they might have multiple top level folders to actually store notes in which they may want to link to.

True. In that case please ignore the second thing I said, but "I think it would be a lot more efficient to scan only directories (use only_dirs = true) and then check if the given note name is in each directory" is still relevant.

shakesbeare: Should be done!

README.md
```diff
@@ -26,6 +26,7 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
   This command has one optional argument: the ID, path, or alias of the note to link to. If not given, the selected text will be used to find the note with a matching ID, path, or alias.
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
+- `:ObsidianFollowLink` to open a note or create a new note in your vault
```
epwalsh, line 29: How about:

```suggestion
- `:ObsidianFollowLink` to follow a note reference under the cursor.
```

lua/obsidian/command.lua
```diff
@@ -294,6 +294,63 @@ command.complete_args = function(client, arg_lead, cmd_line, cursor_pos)
...
+
+  if not note_name:match("/") then
+    scan.scan_dir(vim.fs.normalize(tostring(client.dir)), {
+      hidden = false,
+      add_dirs = false,
+      only_dirs = true,
+      respect_gitignore = true,
+      on_insert = function(entry)
+          note_dir = entry .. "/" .. note_name
+          ok, _ = os.rename(note_dir, note_dir)
```
epwalsh, line 332: What's the rename for?

lua/obsidian/util.lua
epwalsh: Sounds good, I never tested my suggestion so maybe I got something wrong 😉

lua/obsidian/command.lua
shakesbeare: From my searching, lua doesn't have a function to check if a file exists built in and I also couldn't find that operation in the plenary plugin. So, it's basically a hack I found on SO to check if a file exists. `ok` will only be true if the file exists and the rename operation successfully renames the file to itself. 

It's definitely possible I missed something, though, and am happy to change it if there's a clearer way. Otherwise, I'll just add a comment.

epwalsh: Oh, you can use the `Path` object from plenary like this:

```lua
note_dir = Path:new(entry .. "/" .. note_name)
if note_dir:is_dir() then
    ...
```

shakesbeare: Done!

lua/obsidian/command.lua
```diff
@@ -294,6 +294,62 @@ command.complete_args = function(client, arg_lead, cmd_line, cursor_pos)
...
+    scan.scan_dir(vim.fs.normalize(tostring(client.dir)), {
+      hidden = false,
+      add_dirs = false,
+      only_dirs = true,
+      respect_gitignore = true,
+      on_insert = function(entry)
+        note_dir = Path:new(entry .. "/" .. note_name)
+        if note_dir:is_dir() then
+          local _, note = pcall(Note.from_file, entry .. "/" .. note_name, client.dir)
+          table.insert(notes, note.path)
```
epwalsh, line 334: `note` could be `nil` if this fails, so I think you need:

```suggestion
          local ok, note = pcall(Note.from_file, entry .. "/" .. note_name, client.dir)
          if ok then
            table.insert(notes, note.path)
          end
```

epwalsh:
Looks great! I'll test this out right now. Looks like the Style and Lint checks are complaining about some minor things

lua/obsidian/command.lua
```diff
@@ -294,6 +294,62 @@ command.complete_args = function(client, arg_lead, cmd_line, cursor_pos)
...
+
+  if open == nil or close == nil then
+    return nil
+  end
+
+  local note_name = current_line:sub(open + 2, close - 1)
+  local path = client.dir
+
+  if note_name:match "|[^%]]*" then
+    note_name = note_name:sub(1, note_name:find "|")
```
epwalsh, line 314:
```suggestion
    note_name = note_name:sub(1, note_name:find "|" - 1)
```

lua/obsidian/command.lua
```diff
@@ -294,6 +294,62 @@ command.complete_args = function(client, arg_lead, cmd_line, cursor_pos)
...
+      hidden = false,
+      add_dirs = false,
+      only_dirs = true,
+      respect_gitignore = true,
+      on_insert = function(entry)
+        local note_dir = Path:new(entry .. "/" .. note_name)
+        if note_dir:is_dir() then
+          local ok, note = pcall(Note.from_file, entry .. "/" .. note_name, client.dir)
+          if ok then
+            table.insert(notes, note.path)
```
epwalsh, line 334: Oops, I think you need this:
```suggestion
        local note_path = Path:new(entry .. "/" .. note_name)
        if note_dir:is_file() then
          local ok, _ = pcall(Note.from_file, note_path, client.dir)
          if ok then
            table.insert(notes, note_path)
```

lua/obsidian/command.lua
```diff
@@ -294,6 +294,62 @@ command.complete_args = function(client, arg_lead, cmd_line, cursor_pos)
...
+---
+---@param client obsidian.Client
+command.follow = function(client, _)
+  local scan = require "plenary.scandir"
+
+  local open, close = util.cursor_on_markdown_link()
+  local current_line = vim.api.nvim_get_current_line()
+
+  if open == nil or close == nil then
+    return nil
```
epwalsh, line 307:
```suggestion
    echo.err "Cursor is not on a reference!"
    return
```

epwalsh (commented):
I just found a couple more minor issues while testing in my own vault. Fixes below

epwalsh (approved):
LGTM, thanks for the great contribution @shakesbeare!

PR #43, opened 2022-10-31
require yaml directly

ulyssesdotcodes:
Windows wasn't happy with the symbolic link. This should work on all platforms.

Super excited to use this, but my daily driver is Windows. If it's ok I'd like to submit PRs when I find the time to get this working.


epwalsh (approved):
Thanks @ulyssesdotcodes, LGTM

Issue #42, opened 2022-10-31
Why i can't disable frontmatter?


outp1:
### 🐛 Describe the bug

I **want to disable Frontmatter** 'cause it's so useless feature for me. 
But all my attempts lead me to either default frontmatter {id, aliases, tags} or this:
![image](https://user-images.githubusercontent.com/81954928/198962644-a7919953-86b7-4fb3-b7c9-7acedb06fa5c.png)
second case config: 
```lua
note_frontmatter_func = function(note)
    local date = tostring(os.date())
    local out = { }
    if note.metadata ~= nil then
      for k, v in pairs(note.metadata) do
        out[k] = v
      end
    end
    return out
  end,
```

With `note_frontmatter_func = nil,` I see it as intended:
![image](https://user-images.githubusercontent.com/81954928/198963990-c46341e7-5567-4f72-839b-f5e1b9684a10.png)

_I really hope that when you wrote the frontmatter code you took care to disable it_

### Versions

`nvim --version`
NVIM v0.8.0-dev-798-ge6680ea7c
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

`nvim --headless -c 'lua print(require("obsidian").VERSION)' -c q`
1.6.1

cstsunfu:
I just want to disable this too.

epwalsh:
Happy to accept a PR for this. I think the best way to configure this would be to have a flag in the config like `disable_frontmatter` or something. Or if `note_frontmatter_func` returns `nil`, don't add frontmatter.

outp1:
Ok, I'll take care of it

PR #48, opened 2022-11-10
Added configuration option `disable_frontmatter` (outp1#42)

outp1:
This commit will solve this [issue](https://github.com/epwalsh/obsidian.nvim/issues/42).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 tmp.*
 *.sqlite
 doc/tags
+.luarc.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `:ObsidianFollowLink` and companion function `util.cursor_on_markdown_link`
 - Added `:ObsidianLink` and `:ObsidianLinkNew` commands.
+- Added configuration option `disable_frontmatter` for frontmatter disabling
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ require("nvim-treesitter.configs").setup({
 
 #### Customizing the automatically generated YAML frontmatter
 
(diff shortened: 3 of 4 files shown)


epwalsh (approved):
LGTM, thanks @outp1!

saccarosium:
Hi,
It seems that the option doesn't work, on my system.

my config:
```lua
require("obsidian").setup({
  dir = "~/Documents/nextcloud/notes/",
  note_frontmatter_func = nil,
  notes_subdir = "notes",
    completion = {
      nvim_cmp = true, -- if using nvim-cmp, otherwise set to false
    }
  })

  vim.keymap.set("n", "gf", function()
    if require('obsidian').util.cursor_on_markdown_link() then
      return "<cmd>ObsidianFollowLink<CR>"
    else
      return "gf"
   end
  end,
  { noremap = false, expr = true })

vim.keymap.set("n", "<leader>on", ":ObsidianNew<CR>")
vim.keymap.set("n", "<leader>os", ":ObsidianSearch<CR>")
```
Is located under `after/plugin` but I tested also on `plugin`. I don't know what the problem is.

fermifunktion:
Hi,
on my system `disable_frontmatter = true` is also not working within the `require("obsidian").setup()` function:

``` lua
require("obsidian").setup({
  dir = "~/Obsidian",
    local suffix = ""
    if title ~= nil then
    else
      for _ = 1, 4 do
      end
    end
    -- return tostring(os.time()) .. "-" .. suffix
    return title
   end,
    completion = {
    nvim_cmp = true, -- if using nvim-cmp, otherwise set to false
  },
  disable_frontmatter = true,
})
```
I am using `neovim v0.8.3` on `Arch Linux, Kernel 6.1.11`

epwalsh:
Just pushed a fix.

TLSingh1:
Hi, thank you for your work.

Unfortunately, this doesnt disable it for me. I tried with latest commit and the one specified above. Using packer.

```
		use({
			"epwalsh/obsidian.nvim",
                        commit = "3d45970",
			config = function()
				require("obsidian").setup({
					use_advanced_uri = true,
					dir = "~/brain",
					completion = {
						nvim_cmp = true,
					},
                                        disable_frontmatter = true,
				})
			end,
		})

```

PR #56, opened 2022-11-23
Add line-navigation to :ObsidianOpen (AdvancedURI)

gerryhernandez:
This PR adds a thin integration with the Obsidian Advanced URI plugin, allowing :ObsidianOpen to automatically scroll the Obsidian editor to the same line as NeoVim's.

See: [Obsidian Advanced URI](https://github.com/Vinzent03/obsidian-advanced-uri)

Note: scrolling in the Reading viewmode does not currently work (see [this issue](https://github.com/Vinzent03/obsidian-advanced-uri/issues/91)). The Obsidian editor will scroll correctly when it's in edit-mode.

PR #57, opened 2022-11-25
Fix URL encoding for space characters

gerryhernandez:
I ran into this one while using #56 , but this should improve general compatibility for anything that needs URI encoding. Turns out, encoding URIs in Lua is kind of a pain. Didn't know that. But this seems to do the trick for all the edge cases I personally have.

TLDR here is that `%20` is always a safe encoding for a space character, regardless of where it is in the raw URL string, but `+` is interpreted as a literal if it's in a query param value.


epwalsh (approved):
Thanks @gerryhernandez!

PR #67, opened 2022-12-22
Fixed small bug in note.from_lines

s-cassidy:
If a note had frontmatter consisting of exactly one empty field, `yaml.loads` would return that field as a string rather than a table, causing an exception in `note.from_lines`.

### DIFF
--- a/lua/obsidian/note.lua
+++ b/lua/obsidian/note.lua
@@ -209,6 +209,9 @@ note.from_lines = function(lines, path, root)
   if #frontmatter_lines > 0 then
     local frontmatter = table.concat(frontmatter_lines, "\n")
     local ok, data = pcall(yaml.loads, frontmatter)
+    if type(data) == 'string' then
+      data = {data = nil}
+    end
     if ok then
       for k, v in pairs(data) do
         if k == "id" then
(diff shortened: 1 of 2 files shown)


lua/obsidian/note.lua
```diff
@@ -209,6 +209,9 @@ note.from_lines = function(lines, path, root)
   if #frontmatter_lines > 0 then
     local frontmatter = table.concat(frontmatter_lines, "\n")
     local ok, data = pcall(yaml.loads, frontmatter)
+    if type(data) == 'string' then
+      data = {data = nil}
```
epwalsh, line 213: How about just this:
```suggestion
      data = {}
```

epwalsh (commented):
Thanks for the PR @s-cassidy! Just one minor suggestion

s-cassidy:
That's my poor Lua knowledge showing me up as I didn't realize nil table entries effectively don't exist. I've made the change. However, this is closer to what I was thinking
```
if type(data) == 'string' then
  local field = data
  field = field:gsub("^%s*(.-):?%s*$", "%1") 
  data = {}
  data[field] = {}
end
```
This way the following code still gets the name of the empty field (with trailing whitespace and `:` removed) with an empty table as the value.

epwalsh (approved):
Thanks, LGTM

Issue #66, opened 2022-12-19
Find note


joserc87:
### 🚀 The feature, motivation and pitch

First of all, thanks for this marvelous plugin and apologies if the feature is there, but I couldn't find it.

One of the most basic things I do in obsidian is to `<Ctrl-O>` to quickly open a note, if I remember the name. I'm not talking about searching _inside_ the notes, but trying to open a specific note by name, so this is not the same usecase as `:ObsidianSearch`, it's much simpler than that. For example, let's say I just want to quickly read my `obsidian.nvim` cheatsheet. I'd like to run something like `:ObsidianSelectNote` (conveniently mapped to some shortcut) to see a list in Telescope with *only* the name of the notes (so it's a much more reduced list than with `:ObsidianSearch`) and from there I can type the name of the note to open it.

AFAIK there is no such command, but if there is, I'm happy to send a PR to change the README.

### Alternatives

I guess that for the time being `:ObsidianSearch` does the job as well, because it _also_ searches for the filename, but if I already know the name of the note, having to search through the content as well seems inefficient.

epwalsh:
Hey @joserc87, currently Obsidian relegates that functionality to other plugins. Personally I use `fern` in combination with `fzf` via [fern-mapping-fzf.vim](https://github.com/LumaKernel/fern-mapping-fzf.vim). But it would be nice to have something like that within this plugin.

joserc87:
Fair enough. It's just a list of files, so I can hack it myself into telescope. The only benefit of having it inside obsidian.nvim is that I don't have to spread the configuration to multiple places. Since obsidian.nvim "knows" where my notes are, it can list the files. Not a huge deal, but it would be a nice to have.
I'm a complete noob when it comes to lua and neovim plugins, but I could try to give it a go and try to create a PR, because as I said, I love this plugin. Kudos!

epwalsh:
Great, I'd love to see a PR for this if you find the time!

joserc87:
I sent a PR.
> Personally I use fern in combination with fzf via [fern-mapping-fzf.vim](https://github.com/LumaKernel/fern-mapping-fzf.vim).
I'm very interested to know if there is anything in your current workflow that is not in the new command. We should make it at least as useful as your current workflow, so that you can make the switch.

joserc87:
#68 got merged, so I'm happy to resolve this issue :tada:

PR #68, opened 2022-12-26
feat: ObsidianQuickSwitch command

joserc87:
Hi!
As requested in #66, I'd like to quickly open an Obsidian note, so I created a new command for this. Since in the Obsidian App this action (`Ctrl-O`) is called "Open Quick Switcher", I called the command `:ObsidianQuickSwitch`, because `:ObsidianOpenQuickSwitcher` is longer than needed. Of course, I'm open for suggestions.

Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `:ObsidianQuickSwitch` to fuzzy-find a note by name in telescope/fzf _a la_ <C-O> in Obsidian.
 - Added support for [fzf-lua](https://github.com/ibhagwan/fzf-lua) as one of the possible fallbacks for the `:ObsidianSearch` command.
 - Added `:ObsidianFollowLink` and companion function `util.cursor_on_markdown_link`
 - Added `:ObsidianLink` and `:ObsidianLinkNew` commands.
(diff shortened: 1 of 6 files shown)


joserc87:
Any feedback on the code is very much welcome, because I'm quite new to lua (specifically, I don't know if there exists a function already for converting a list of strings to a single string; if there is I'm happy to remove the function `table_params_to_str`).

epwalsh (commented):
Thanks @joserc87, this looks great! Looks like there's just some minor lint/style issues that CI is complaining about. Once you fix those I'll hit the merge button!

joserc87:
All the CI issues should be fixed now @epwalsh .

Since I started this PR a couple of weeks ago, other PRs got merged in, in particular to add support for `fzf-lua`, so I had to deal with some merge conflicts. The command that I added doesn't take fzf-lua into account, so it doesn't work with it, it requires either FZF or Telescope. At first I was thinking to add support for `fzf-lua`, but I think I should do something more clever than copying the code with the `if has_fzf_lua` etc, to keep the code maintainable...

I can add the support for fzf-lua  as part of a separate PR, so that we can merge this one and I don't get ever more merge conflicts... :) Do you agree?

epwalsh (approved):
> I can add the support for fzf-lua as part of a separate PR, so that we can merge this one and I don't get ever more merge conflicts... :) Do you agree?

Sounds good to me! FYI @saccarosium

PR #69, opened 2022-12-31
Adding support for fzf-lua in search command

saccarosium:
Currently the only fuzzy finders implemented are telescope.nvim and fzf.vim. 
I use [fzf-lua](https://github.com/ibhagwan/fzf-lua), which is a lua based fzf plugin, that work like telescope but uses fzf in the background for better performance and, in my opinion, better fuzzy algorithm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Plug 'hrsh7th/nvim-cmp'
 Plug 'junegunn/fzf', { 'do': { -> fzf#install() } }
 Plug 'junegunn/fzf.vim'
 
+" (optional) another alternative for the :ObsidianSearch command:
+Plug 'ibhagwan/fzf-lua'
+
 " (optional) for :ObsidianSearch command if you prefer this over fzf.vim:
 Plug 'nvim-telescope/telescope.nvim'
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,6 +51,9 @@ Plug 'hrsh7th/nvim-cmp'
 Plug 'junegunn/fzf', { 'do': { -> fzf#install() } }
 Plug 'junegunn/fzf.vim'
 
+" (optional) for :ObsidianSearch command unless you use telescope:
```
epwalsh, line 54:
```suggestion
" (optional) another alternative for the :ObsidianSearch command:
```

epwalsh (commented):
Didn't know about `fzf-lua`! Thanks for this, just one comment

saccarosium:
Fixed comment and added fix in search.

epwalsh:
Can you just add an entry to the CHANGELOG under `### Added` that says something like "Added support for `fzf-lua`."

saccarosium:
> Can you just add an entry to the CHANGELOG under `### Added` that says something like "Added support for `fzf-lua`."

Done!

epwalsh (approved):
LGTM! Thank you!!

PR #73, opened 2023-01-06
docs: fix undefined global `util`

kohane27:
Thank you for this awesome plugin!


epwalsh (approved):
Good catch! Thanks @kohane27

PR #76, opened 2023-01-07
Fix navigating to notes in the vault's root directory

gerryhernandez:
I came across this paper cut today. Turns out, Plenary doesn't consider the actual directory it's scanning, only its children. The net effect is that you'd never come across this bug if you only use notes within a subdirectory. But I started using top-level notes today, so I came across this bug.

The official Obsidian app does match this behavior.


epwalsh (approved):
Thanks @gerryhernandez!

PR #82, opened 2023-01-12
Adding support for fzf-lua in the quickswitch command

saccarosium:
As mentioned in #68

### DIFF
--- a/lua/obsidian/command.lua
+++ b/lua/obsidian/command.lua
@@ -224,6 +224,15 @@ command.quick_switch = function(client, data)
     return
   end
 
+  local has_fzf_lua, fzf_lua = pcall(require, "fzf-lua")
+
+  if has_fzf_lua then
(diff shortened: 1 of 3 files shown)


lua/obsidian/command.lua
```diff
@@ -1,5 +1,3 @@
-local Path = require "plenary.path"
-local Job = require "plenary.job"
```
epwalsh, line 2: I'm guessing you didn't mean to remove these?

epwalsh (changes requested):
LGTM, just one issue

lua/obsidian/command.lua
saccarosium: Strange I don't really know what happened

saccarosium:
Should be good

epwalsh (approved):
Thank you!!

PR #86, opened 2023-01-23
Save new notes created from auto complete in the current folder

rag-hav:
Currently new notes that are created through auto complete from nvim-cmp are saved in the notes_subdir directory.

I like to organize my notes in folders by topic and by saving new notes in the current directory would save a lot of effort


lua/obsidian/init.lua
```diff
@@ -197,7 +197,7 @@ client.new_note = function(self, title, id)
 
   -- Get path.
   ---@type Path
-  local path = Path:new(self.dir)
+  local path = fldr == nil and Path:new(self.dir) or Path:new(fldr)
```
epwalsh, line 200: What happens if `opts.notes_subdir` is set, and `fldr` is equal to that directory? I think we'd want to check for that.

For example, my vault is at `~/epwalsh-notes`, and I have `opts.notes_subdir` set to `notes/`. So if I'm editing a file in `~/epwalsh-notes/notes/` and autocomplete a new note, I think `fldr` here would be `~/epwalsh-notes/notes/`, so below we'd set `path` to `~/epwalsh-notes/notes/notes/`.

epwalsh (changes requested):
Thanks for the PR @rag-hav! This seems like a good feature, I just have one concern about the implementation.

lua/obsidian/init.lua
rag-hav: Yes, I should have thought of this one. Changed it so if the dir is given `notes_subdir` is ignored.

rag-hav:
Hi,

I am closing this pull request since I discovered that an easier way of achieving this functionality is through the `note_id_func` function.

```
				local cd = vim.fn.expand("%:~:h") .. '/'
				cd = string.gsub(cd, notesDir, '')
				return cd .. suffix

```
Adding this to that function lets me achieve the same functionality, and I like it more this way since now the suggestion also includes the folder name, and the given id is consistent with when a notes file is created manually.

rag-hav:
The folder name will clutter the notes, so reopening this :sweat_smile:

epwalsh (approved):
LGTM!

PR #91, opened 2023-02-07
Add `:ObsidianRename` command

epwalsh:
The idea is to be able to rename notes as easily as you can from the Obsidian app. The challenge here is that we need to update all backlinks to the old note to use the new file name/ID. To avoid issues of modifying files that might currently be loaded into a buffer, we have to do this in two steps:

- [x] Update all backlinks to the note that are in open buffers of other notes.
- [ ] Update all backlinks to the note in all other notes in the vault that are not currently loaded in a buffer. This will require external dependencies like `find` and `sed`, or if we want to stick with tools from the Rust ecosystem: `fd` and `sd`. Personally I'd prefer the latter since we wouldn't have to account for differences between the Mac vs GNU vs Windows(?) versions of these tools. The downside of course is that it adds more things people need to install.

Other requirements:

- [x] We should be able to update markdown-style links in addition to wikilinks.
- [x] When you select this command, the arguments should auto-populate with the current name of the note.
    ⚠️ NOTE: this works now, but the caveat is you have to TAB after selecting the command to trigger the autocomplete. I'm not sure if there's a way to avoid that.

### DIFF
--- a/lua/obsidian/command.lua
+++ b/lua/obsidian/command.lua
@@ -329,7 +329,7 @@ command.link = function(client, data)
   vim.api.nvim_buf_set_lines(0, csrow - 1, csrow, false, { line })
 end
 
-command.complete_args = function(client, _, cmd_line, _)
+command.complete_args_search = function(client, _, cmd_line, _)
   local search
   local cmd_arg, _ = util.strip(string.gsub(cmd_line, "^.*Obsidian[A-Za-z0-9]+", ""))
   if string.len(cmd_arg) > 0 then
@@ -378,26 +378,17 @@ end
 ---
 ---@param client obsidian.Client
 command.follow = function(client, _)
-  local open, close = util.cursor_on_markdown_link()
-  local current_line = vim.api.nvim_get_current_line()
-
-  if open == nil or close == nil then
+  local note_name, note_path = util.cursor_link()
+  if note_name == nil or note_path == nil then
     echo.err "Cursor is not on a reference!"
     return
   end
 
-  local note_name = current_line:sub(open + 2, close - 1)
-  local note_file_name = note_name
-
-  if note_file_name:match "|[^%]]*" then
-    note_file_name = note_file_name:sub(1, note_file_name:find "|" - 1)
+  if note_path:match "%.md" == nil then
+    note_path = note_path .. ".md"
   end
 
-  if not note_file_name:match "%.md" then
-    note_file_name = note_file_name .. ".md"
-  end
-
-  local notes = util.find_note(client.dir, note_file_name)
+  local notes = util.find_note(client.dir, note_path)
 
   if #notes < 1 then
     command.new(client, { args = note_name })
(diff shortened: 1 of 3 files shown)


gerryhernandez:
So I'm not a Lua or NeoVim plugin dev, but I can fake it sometimes. I have a dumb question: why not just write all open (relevant) buffers, then go ahead and manipulate all the backlinks (on disk), then reload the already open buffers? My primary concern is that if the entire operation isn't done on disk, the notebook could be left in an inconsistent state. For example, if a an open buffer isn't saved by the user, the backlink will become a dangling link.

As far as using GNU-style tools versus the Rust replacements, what kind of performance penalty are we talking if we do the entire thing inside NeoVim using hidden buffers? The advantage here would be consistency: one code path for all backlink manipulations. If this is a severe performance penalty, then I'm with you on choosing `fd`, `sd`, etc. There could also be an upgrade path here.

Thanks for getting this started! The ecosystem is getting pretty close to working independently of Obsidian for simple use cases, which mine are, so I very much appreciate your time spent on this. Let me know how I can help. :)

epwalsh:
> why not just write all open (relevant) buffers, then go ahead and manipulate all the backlinks (on disk), then reload the already open buffers?

That's not a bad idea. I steered away from that originally because I think it would be unexpected from a user perspective to have all of your buffers written as the result of running a command. But if we're transparent about the consequences of running the command, it's probably okay.

> what kind of performance penalty are we talking if we do the entire thing inside NeoVim using hidden buffers?

I suspect it would be pretty bad for big vaults, unless there's a good multi-threaded approach. But reading `:help threading`, it doesn't sound promising: "Note that every thread gets its own separate lua interpreter state, with no access to lua globals in the main thread. Neither can the state of the editor (buffers, windows, etc) be directly accessed from threads."

gerryhernandez:
I think it would be safe and sane to prompt the user with a warning about the operation, requiring them to say “yes,” where “no” is default. But also allow the user to opt-out if the prompt via a config option, of course.

epwalsh:
> I think it would be safe and sane to prompt the user with a warning about the operation, requiring them to say “yes,” where “no” is default. But also allow the user to opt-out if the prompt via a config option, of course.

Yeup, I like that. And the prompt could say how to opt out via the config.

> A single code-path is still possible with sd and fd if the whole thing is done on disk, which might be worth the consistency and maintenance gains.

👍

A:
Hi, in my own try to implement obsidian behavior, I have this feature:

```lua
local function get_cur_note_name()
  local file = api.nvim_buf_get_name(0)
  return get_basename(file)
end

local function remove_cur_note()
  local file = api.nvim_buf_get_name(0)
  cmd(':!rm "' .. file .. '"')
end

local function rename()
  local basename = get_cur_note_name()

  if (basename == nil) then
    return
  end

  ui.input({ prompt = 'Enter a new name for ' .. basename ..': ' }, function(input)
    cmd(':w ' .. input .. '.md')
    cmd('!find . -name "*.md" -type f -exec sed -i "" "s/\\[\\[' .. basename .. '\\([ \\t]*\\|[ \\t]*.*\\)\\{0,1\\}\\]\\]/[[' .. input ..'\\1]]/g" {} +')
    remove_cur_note()
  end)
end

```

Just in case if it can help to get this feature merged, but I see it's almost done on your side :)

epwalsh:
Closing in favor of #232.

PR #98, opened 2023-02-23
Added support for templates

s-cassidy:
I have added a new command `:ObsidianTemplate` that allows inserting of a template from the templates folder. The template is selected using telescope and supports the Obsidian features of substituting `{{date}}`, `{{time}}` and `{{title}}`. The templates folder and date/time formats are configurable via a `templates` table passed to `setup()`.

Unsupported so far is the ability to substitute "one-off" formats for date and time e.g. `{{date:YYYY-MM-DD}}` (in Obsidian's syntax). It would also be nice to read the `templates.json` file directly from `.obsidian` to get the configuration -- I've made some moves toward this but it's not not done yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+
+### Added
+
+- Added `:ObsidianTemplate` to insert a template, configurable using a `templates` table passed to `setup()`.
+
 ## [v1.8.0](https://github.com/epwalsh/obsidian.nvim/releases/tag/v1.8.0) - 2023-02-16
 
 ### Changed
 
 - [`lua-yaml`](https://github.com/exosite/lua-yaml) no-longer bundled as a git submodule. Code from that project has been copied and modified into it's own Lua submodule of `obsidian`.
 - (BREAKING) 'nvim-lua/plenary.nvim' is no-longer bundled, so must be explicitly installed (e.g. Plug 'nvim-lua/plenary.nvim' in your `init.nvim`).
-
-### Added
-
-- Added `:ObsidianTemplate` to insert a template, configurable using a `templates` table passed to `setup()`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [`lua-yaml`](https://github.com/exosite/lua-yaml) no-longer bundled as a git submodule. Code from that project has been copied and modified into it's own Lua submodule of `obsidian`.
 - (BREAKING) 'nvim-lua/plenary.nvim' is no-longer bundled, so must be explicitly installed (e.g. Plug 'nvim-lua/plenary.nvim' in your `init.nvim`).
 
+### Added
+
+- Added `:ObsidianTemplate` to insert a template, configurable using a `templates` table passed to `setup()`.
+
 ### Fixed
 
 - Fixed a bug where creating a new note with `nvim-cmp` completion
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
 - `:ObsidianFollowLink` to follow a note reference under the cursor.
+- `:ObsidianTemplates` to insert a template from the templates folder, selecting from a list using [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).
 
 ## Setup
 
@@ -164,6 +165,21 @@ require("obsidian").setup({
   notes_subdir = "notes",
   daily_notes = {
     folder = "notes/dailies",
+  }
+})
+```
+
+#### Templates support
+
+To insert a template, run the command `:ObsidianTemplates`. This will open [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim) and allow you to select a template from the templates folder. Select a template and hit `<CR>` to insert. Substitution of `{{date}}`, `{{time}}`, and `{{title}}` is supported. 
+
+```lua
+require("obsidian").setup({
+  dir = "~/my-vault",
+  templates = {
+      subdir = "my-templates-folder",
+      date_format = "%Y-%m-%d-%a",
+      time_format = "%H:%M"
   }
 })
 ```
--- a/README.md
+++ b/README.md
@@ -185,6 +185,7 @@ require("obsidian").setup({
   }
 })
 ```
+
 and the file `~/my-vault/my-templates-folder/note template.md`:
 
 ```markdown
@@ -195,8 +196,10 @@ Date created: {{date}}
 creating the note `Configuring Neovim.md` and executing `:ObsidianTemplate` will insert
 ```markdown
 # Configuring Neovim
+
 Date created: 2023-03-01-Wed
 ```
+
 above the cursor position.
 
 #### Using nvim-treesitter
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
 - `:ObsidianFollowLink` to follow a note reference under the cursor.
-- `:ObsidianTemplates` to insert a template from the templates folder, selecting from a list using [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [`lua-yaml`](https://github.com/exosite/lua-yaml) no-longer bundled as a git submodule. Code from that project has been copied and modified into it's own Lua submodule of `obsidian`.
 - (BREAKING) 'nvim-lua/plenary.nvim' is no-longer bundled, so must be explicitly installed (e.g. Plug 'nvim-lua/plenary.nvim' in your `init.nvim`).
 
+### Added
+
+- Added `:ObsidianTemplate` to insert a template, configurable using a `templates` table passed to `setup()`.
+
```
epwalsh, line 20: Can you move this up to the "Unreleased" section?

README.md
```diff
@@ -168,6 +169,21 @@ require("obsidian").setup({
 })
 ```
 
+#### Templates support
+
+To insert a template, run the command `:ObsidianTemplates`. This will open [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim) and allow you to select a template from the templates folder. Select a template and hit `<CR>` to insert. Substitution of `{{date}}`, `{{time}}`, and `{{title}}` is supported.
```
epwalsh, line 174: Could you add an example template as well?

lua/obsidian/command.lua
```diff
@@ -216,6 +216,77 @@ command.search = function(client, data)
...
+          apply_template(template[1])
+          require("telescope.actions").close(prompt_bufnr)
+        end)
+        return true
+      end,
+    }
+    require("telescope.builtin").find_files(opts)
+  end
+  choose_template()
+  return
```
epwalsh, line 286: Could you add support for `fzf` as well if it's not too difficult? See the `ObsidianSearch` and `ObsidianQuickSwitch` commands for an example

epwalsh (changes requested):
Hi @s-cassidy, thanks for this! I just have a few suggestions

lua/obsidian/command.lua
s-cassidy: I have made these changes and have added support for `fzf-lua` and `fzf.vim`.

For `fzf.vim` the only way I could find to get the output from `fzf.vim` and into my lua function was to register and then shortly after remove a new command for `fzf` to send its output to (I don't know vimscript so there's probably a better way).

Finally, there is some weirdness in `fzf-lua` in which the file name it passes comes with 6 characters at the start that are not part of the filename and render as two blank spaces -- I'll open an issue with `fzf-lua`.

epwalsh: Thanks @s-cassidy, I'll try it out with fzf today and see if I can figure out what's happening there

README.md
```diff
@@ -29,6 +29,7 @@ Built for people who love the concept of Obsidian -- a simple, markdown-based no
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
   This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
 - `:ObsidianFollowLink` to follow a note reference under the cursor.
+- `:ObsidianTemplate` to insert a template from the templates folder, selecting from a list using [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).
```
epwalsh, line 32:
```suggestion
- `:ObsidianTemplate` to insert a template from the templates folder, selecting from a list using [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim) or one of the `fzf` alternatives.
```

README.md
```diff
@@ -168,6 +169,44 @@ require("obsidian").setup({
 })
 ```
 
+#### Templates support
+
+To insert a template, run the command `:ObsidianTemplate`. This will open [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim) and allow you to select a template from the templates folder. Select a template and hit `<CR>` to insert. Substitution of `{{date}}`, `{{time}}`, and `{{title}}` is supported.
```
epwalsh, line 174:
```suggestion
To insert a template, run the command `:ObsidianTemplate`. This will open [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim) or one of the `fzf` alternatives and allow you to select a template from the templates folder. Select a template and hit `<CR>` to insert. Substitution of `{{date}}`, `{{time}}`, and `{{title}}` is supported. 
```

lua/obsidian/command.lua
```diff
@@ -216,6 +216,123 @@ command.search = function(client, data)
...
+    fzf_lua.files {
+      cmd = cmd,
+      cwd = tostring(templates_dir),
+      file_icons = false,
+      actions = { ["default"] =
+      function(entry)
+        -- for some reason fzf-lua passes the filename with 6 characters
+        -- at the start that appear on screen as 2 whitespace characters
+        -- so we need to start on the 7th character
+        local template = entry[1]:sub(7)
```
epwalsh, line 302: If those are all whitespace characters it would probably be more robust to trim off the whitespace using gsub or match. Here are some example implementations: http://lua-users.org/wiki/StringTrim

README.md
```diff
@@ -168,6 +169,44 @@ require("obsidian").setup({
...
+  }
+})
+```
+and the file `~/my-vault/my-templates-folder/note template.md`:
+
+```markdown
+---
+aliases = []
+---
+
```
epwalsh, line 194: By default Obsidian handles the frontmatter by itself, so I think we keep this example consistent with the default settings and omit frontmatter in the template (otherwise you'd end up with duplicate frontmatter).
```suggestion
```

README.md
```diff
@@ -168,6 +169,44 @@ require("obsidian").setup({
...
+# {{title}}
+Date created: {{date}}
+```
+
+creating the note `Configuring Neovim.md` and executing `:ObsidianTemplate` will insert
+```markdown
+---
+aliases = []
+---
+
```
epwalsh, line 204: Same as above
```suggestion
```

README.md
```diff
@@ -168,6 +169,44 @@ require("obsidian").setup({
...
+# Configuring Neovim
+Date created: 2023-03-01-Wed
+```
+above the cursor position.
```
epwalsh, line 208: I think it's more intuitive to insert *below* the cursor position. Is there a reason you chose to insert above?

lua/obsidian/command.lua
```diff
@@ -216,6 +216,123 @@ command.search = function(client, data)
...
+
+  local apply_template = function(name)
+    local template_path = Path:new(templates_dir / name)
+    local date_format = client.opts.templates.date_format or "%Y-%m-%d"
+    local time_format = client.opts.templates.time_format or "%H:%M"
+    local date = tostring(os.date(date_format))
+    local time = tostring(os.date(time_format))
+    local current_fp = vim.api.nvim_buf_get_name(buf)
+    local _, _, title = string.find(vim.fs.normalize(current_fp), ".*/(.-)%.md")
+    title = title or vim.fs.basename(current_fp)
```
epwalsh, line 249: It would be more robust to use the `Note` class here:
```suggestion
    local title = Note.from_buffer(buf, client.dir):display_name()
```

epwalsh (changes requested):
Hey @s-cassidy, I tested it with all three options (`telescope`, `fzf-lua`, `fzf`) and they all work great! Awesome job with that. I left a few suggestions to make things more robust. My one other concern is with the behavior of inserting *above* the cursor, which seems less intuitive than inserting below the cursor. If you feel strongly about that, we could just add another configuration option that lets the user decide.

README.md
s-cassidy: Agreed - I overlooked this as I disabled that feature

lua/obsidian/command.lua
s-cassidy: Tried these, doesn't work. It's a very strange behaviour from fzf-lua I think

README.md
epwalsh: > Obsidian.md's default behaviour and it inserts the template at (before, if in using obsidian's vim mode in normal mode) the current cursor position, not linewise, which may be another option.

Actually I like that even better!

lua/obsidian/command.lua
epwalsh: Weird.. I guess those characters aren't typical whitespace?

README.md
s-cassidy: I believe this works as in Obsidian.md now

lua/obsidian/command.lua
s-cassidy: Yup, I'm also struggling to reproduce it in another script to try and isolate the problem.
```lua
local fzf_lua = require("fzf-lua")
fzf_lua.files {
  cmd = "find -name '*.lua'",
  file_icons = false,
  git_icons = false,
  actions = { ["default"] =
  function(entry)
    print(#entry[1])
  end
  }
}
```
Sourcing this and picking a file is printing the correct number of characters, no weird extra spaces.

README.md
epwalsh: Nice!

epwalsh (approved):
Sorry I was slow to get back to you. Looks great! Thanks a lot for this. If you figure out the weird spacing issue please make a PR, but since it seems to work fine I'm not too worried about it.

Issue #50, opened 2022-11-18
[Feature] Follow header links


cmvanb:
### 🚀 The feature, motivation and pitch

I use header links in Obsidian like these:

```
[[#Some header in my note]]

...

### Some header in my note
```

In Obsidian, clicking the header link scrolls down to the linked header in the document.

It would be useful if obsidian.nvim could do the same with `:ObsidianFollowLink`.

### Alternatives

The alternative is scrolling down manually.

epwalsh:
Agreed, this would be great. Contributions welcome!

swaits:
+1 on this.

Any pointers on where to start look at implementation? I'm glad to look at it, but if you have any ideas on how you want it done, @epwalsh, I'll use that as guidance.

epwalsh:
Hey @swaits, it looks like we'd start by checking for a "#" at the beginning of the link name, somewhere after [this point](https://github.com/epwalsh/obsidian.nvim/blob/0418505ff7d1e06460450a02b9209d03f4c2a50d/lua/obsidian/command.lua#L534) in the `command.follow` function. The tricky part would then be iterating over lines in the buffer to find the matching header. It would be good to split that functionality into a util function (`lua/obsidian/util.lua`).

mscott99:
I use links of the form [[file_name#header_section]], and clicking on them creates a wrong file "file_name#header.md" that I then have to delete later. A quick fix would be to simply ignore anything after the '#' and treat it as a normal link; that would already be a big improvement.

epwalsh:
@mscott99 quick fix in #152.

MohamedOsman1998:
I made a comment in #152 and ill make it here as well, but the fix in #152 doesnt fix links that use ^. i.e. block links. A fix for that would be great as well!

videbar:
Hi @epwalsh, I'd like to do some work on this.

I've used tree-sitter to prototype two functions:

- `get_header_pos()`: Takes a header title and find the position of the section in the file. I think this could be used inside [`command.follow`](https://github.com/epwalsh/obsidian.nvim/blob/0418505ff7d1e06460450a02b9209d03f4c2a50d/lua/obsidian/command.lua#L517) to jump to the right position in the file being opened (e.g., using something like `:edit +<line number> <filename>`). If the reference is in the current buffer, only the cursor needs to move.

[...]

PR #99, opened 2023-03-13
feat: follow header links (#50)

Tejada-Omar:
Allows user to link to headers in the general form `[[note#header]]`

Also accepts:

- `[[#header]]` to link to a header in the same file
- `[[#header1#header2]]` to indicate hierarchy

### DIFF
--- a/lua/obsidian/note.lua
+++ b/lua/obsidian/note.lua
@@ -408,17 +408,4 @@ note.save = function(self, path, insert_frontmatter)
   return lines
 end
 
---Jump to a series of nested headers
---Used with :ObsidianFollowLink to follow links in the form [[note#header]]
---Also accepts [[#header]] or [[note#header1#header2]] to indicate hierarchy
---
----@param headers table
-note.jump_to_header = function(headers)
-  -- Begin search at beginning of note
-  vim.fn["cursor"](1, 1)
-  for _, header in ipairs(headers) do
-    vim.cmd("/^#\\+ " .. header .. "$")
-  end
-end
-
 return note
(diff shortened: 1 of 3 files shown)


Tejada-Omar:
Note that if you're using completion it will still offer to create a new note with the header's name, I'll try and fix this if the rest looks good

lua/obsidian/note.lua
```diff
@@ -408,4 +408,17 @@ note.save = function(self, path, insert_frontmatter)
   return lines
 end
 
+--Jump to a series of nested headers
+--Used with :ObsidianFollowLink to follow links in the form [[note#header]]
+--Also accepts [[#header]] or [[note#header1#header2]] to indicate hierarchy
+--
+---@param headers table
+note.jump_to_header = function(headers)
```
epwalsh, line 416: I think it would make more sense to put this into `lua/obsidian/util.lua`. The functions in this file are meant to be instance methods on `Note` objects.

epwalsh (changes requested):
This is pretty cool! Just a minor comment

lua/obsidian/note.lua
Tejada-Omar: On second thought I agree, have pushed a change moving this to `util.lua`

epwalsh: Thanks! Can you fix the issue with new note completion and then we'll get this merged?

epwalsh:
This is stale, so I'm going to close.

one2three4f:
Somewhat unrelated but https://github.com/artempyanykh/marksman is able to full in the gap for this feature, I think itsynergizes quite well with Obsidian.nvim?

Issue #41, opened 2022-10-21
Add a short intro Video/GIF showing how the plugin can be utilised properly. 


DMunkei:
### 🚀 The feature, motivation and pitch

Just a small intro video where different scenarios are showcased would be helpful.

epwalsh:
That would definitely be good to have.

DMunkei:
Yea I looked at it and I didn't know how to use it properly haha.

epwalsh:
Totally understandable. It started off as something I was just building for myself and then (to my surprise) a lot of other people started using it, so it's grown a lot since then and now has a lot of features that aren't completely transparent. Can I ask what your main point of confusion was? Was it how to use the commands or autocompletion or just getting it set up properly?

DMunkei:
Getting it set up in general. I got it working, but didn't understand how to use it.

aniketgm:
Anyone on it ? I came across this plugin and wanted to try it out by referring some video/gif. Instead found an issue, LOL.
(I just started using Obsidian UI App) @epwalsh, Videos can be created later, Or I can help make some too, once I become aware of how to use it. Meanwhile can we have like a simple gif in the README. It can help folks get an overview and to get started. What do you think ?

DMunkei:
Haven't done any videos myself, also found myself not using it because I couldn't wrap my head around it. So I'm just waiting for a video/gif. Sounds like a nice idea to have something like that to pull people in.

epwalsh:
This is in the README now.

PR #108, opened 2023-03-23
Add short demo for new users

shakesbeare:
Adds demonstration video as requested in #41. 

I had a guess at the commands I thought would be most commonly used to show off in a demo that's just over 30s.


epwalsh (approved):
This is great! Thank you

vimpostor:
Looks like you uploaded the demo video to a non-permanent address, because the video in the README is broken.
I also don't see the video in your fork's commit history anymore, which is why Github may have garbage collected the video.

epwalsh:
Hm, it looks like something weird has happened. It doesn't render like it used to for me.

epwalsh:
I think I may have fixed it.

Issue #107, opened 2023-03-23
`ObsidianOpen` opens file as if vault directory is the MacOS root directory if vault dir is behind symbolic link


shakesbeare:
### 🐛 Describe the bug

If obsidian vault is configured to use a directory behind a symlink, using `:ObsidianOpen` while in a note which already exists opens a blank file from the vault directory as if it the vault directory were the root of the entire file tree. 

Example: 
`/path/to/vault/Notes/my_note.md`
Obsidian Opens: `/path/to/vault/path/to/vault/again/Notes/my_note.md`

For completeness, here is a real example using my own paths:
`/Users/username/Library/CloudStorage/Dropbox/Documents/0-obsidian-notes/MUS-490 Landing Page.md`
Obsidian Opens: `/Users/username/Library/CloudStorage/Dropbox/Documents/0-obsidian-notes/Users/my-name/Library/CloudStorage/Dropbox/Documents/0-obsidian-notes/MUS-490 Landing Page.md`

Note that this error occurs regardless of if the note is in a subdirectory of the vault or note. 

Working config:
```lua
require('obsidian').setup {
    dir = "/Users/username/Library/CloudStorage/Dropbox/Documents/0-obsidian-notes",
    notes_subdir = "Notes",
    use_advanced_uri = true,
    completion = {
        nvim_cmp = true,
    },
    templates = {
        subdir = "0-templates"
    }
}
```

Bad config:
```lua
require('obsidian').setup {
    dir = "~/Dropbox/Documents/0-obsidian-notes",
    notes_subdir = "Notes",
    use_advanced_uri = true,
    completion = {
        nvim_cmp = true,
    },
    templates = {
        subdir = "0-templates"
    }
}
```

For what it's worth, it seems to work completely properly as long as I removed the link dir from my config and instead used the absolute dir, but this is absolutely not expected behavior so I figured it's worth documenting at the very least. I can't say for sure whether this happens on other operating systems. 

Seems I also have some error in some kind of python-related thing going on that I just discovered as I was writing this up. Doesn't seem related, but I also don't know what to do about it. Happy to mess with it if you think it's relevant. I haven't noticed any other odd behavior besides in any of my nvim experience besides #106. I did go through and test to see if that issue is still present with the alternate config and it indeed is. 

### Versions

```
NVIM v0.9.0-dev-1265+g410dc3349
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #113, opened 2023-03-28
Fix for `ObsidianOpen` not working with links

shakesbeare:
Potential fix for #107 

### Summary 

This change modifies the `ObsidianOpen` command to additionally check if Plenary's `make_relative` method fails by attempting to find the vault root directory in the path after the call to `make_relative`. If the vault root directory is found, the new implementation simply substitutes out everything up to and including the root directory in an attempt to make the path work.

### Limitations

- This method will not work if the link which the vault is behind is directly to the vault *and* the name of the link does not match the name of the vault. In this case, both `make_relative` and my proposed fix will fail. To fix this would require actually detecting if the vault is behind a link and following it in order to initialize the vault path as its absolute version from the get go. The only way I could find to accomplish this was to use an external library for Lua, which I figured might not be ideal to require as a dependency for what is really a bit of a niche bug. 
- This will also fail if the user happens to keep their note in a subfolder of their vault which shares the same name as the vault itself. This seemed like a more unlikely scenario than someone wanting their vault directory to be behind a link, which would justify the change.
- I believe I have accounted for Windows paths as well, though I imagine that links will look and function differently in Windows. This change may not work for Windows users as a result. That said, it shouldn't break any existing instances that were already working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `nvim-cmp` completion for notes that have no `aliases` specified.
 - `nvim-cmp` completion will search based on file names now too, not just contents.
 - Fixed bug when `nvim-cmp` is not installed.
+- Workaround error which prevented users from using `ObsidianOpen` when vault path was configured behind a link
 
 ## [v1.8.0](https://github.com/epwalsh/obsidian.nvim/releases/tag/v1.8.0) - 2023-02-16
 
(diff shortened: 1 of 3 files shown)


epwalsh (commented):
Thanks for working on this @shakesbeare! Could you just add some comments in the code to explain what's going on and why we have to do that?

epwalsh:
> I was toying with the idea of mentioning the issue somewhere in the README, but wasn't sure how to phrase it given that it should work properly for most everyone now.

I like the idea of adding a "## Known issues" section, and I think saying something like "this is edge case that won't affect most people" is fine.

epwalsh (approved):
LGTM!

I'll hold off on merging until you let me know if you want to add something to the README about known issues.

shakesbeare:
@epwalsh Went ahead and added the README section

PR #123, opened 2023-04-18
feat: detect and handle urls in :ObsidianFollowLink

pbogut:
Following on #122 .

I have implemented URL detection in `:ObsidianFollowLink` and added a default behavior to ignore URLs to prevent the creation of new files for external links (e.g., 'https://example.com/'). 

Additionally, I have added a new configuration option, `follow_url_func`, which allows users to define their own custom behavior when following URLs. This enables users to have more control over what happens when they follow a link, while avoiding the need for `:ObsidianFollowLink` to make assumptions or guesses about user preferences, and handling system differences.


doc/obsidian.txt
```diff
@@ -343,6 +343,26 @@ The other benefit of doing this is that it will now work even if your cursor is
 on the enclosing brackets (`[[` or `]]`) or the alias part of a reference (the
 part after `|`).
 
+CUSTOMIZING THE BEHAVIOR OF FOLLOWING URL LINKS
```
epwalsh, line 346: This doc file is actually generated automatically from the readme, so could you move this to readme instead?

lua/obsidian/command.lua
```diff
@@ -538,6 +538,14 @@ command.follow = function(client, _)
     note_file_name = note_file_name:sub(1, note_file_name:find "|" - 1)
   end
 
+  if note_file_name:match "^[%a%d]*%:%/%/" then
+    if client.opts.follow_url_func ~= nil then
+      client.opts.follow_url_func(note_file_name)
+    end
+    -- early break as we dont want to open urls as notes
```
epwalsh, line 545: How about a warning instead?
```suggestion
    else
      echo.warn "This looks like a URL. You can customize the behavior of URLs with the 'follow_url_func' option."
    end
```

epwalsh (commented):
Two minor comments otherwise looks good!

lua/obsidian/command.lua
pbogut: Oh, yeah, that's surely good idea. Will update

Issue #129, opened 2023-05-03
Allow configuring the finder in the plugin setup


idanarye:
### 🚀 The feature, motivation and pitch

`:ObsidianSearch` is trying - in that order - Telescope, fzf-lua, and fzf.vim:

```lua
command.search = function(client, data)
  local base_cmd = vim.tbl_flatten { util.SEARCH_CMD, { "--smart-case", "--column", "--line-number", "--no-heading" } }

  local has_telescope, telescope = pcall(require, "telescope.builtin")

  if has_telescope then
    -- ...
    return
  end

  local has_fzf_lua, fzf_lua = pcall(require, "fzf-lua")

  if has_fzf_lua then
    -- ...
    return
  end

  -- Fall back to trying with fzf.vim
  local has_fzf, _ = pcall(function()
    -- ...
  end)

  if not has_fzf then
    echo.err "Either telescope.nvim, fzf-lua or fzf.vim is required for :ObsidianSearch command"
  end
end
```

I prefer fzf-lua, but because Telescope is the standard I have it installed for plugins that don't support anything else, so Obsidian will use Telescope on my machine because it's the first in the tryout order.

I would like an option to configure, with a setting, which option `ObsidianSearch` will use. Something like:

```lua
require'obsidian'.setup {
    finder = 'fzf-lua',
}
```

epwalsh:
Hey @idanarye, definitely makes sense! Happy to review a PR here.

epwalsh:
Closed as completed.

PR #130, opened 2023-05-07
feat: add `finder` option for choosing a preferred finder backend (#129)

idanarye:


doc/obsidian.txt
```diff
@@ -180,6 +180,14 @@ Using `lazy.nvim` <https://github.com/folke/lazy.nvim>, for example:
...
         -- Optional, set to true to force ':ObsidianOpen' to bring the app to the foreground.
         open_app_foreground = false,
+
+	-- Optional, by default commands like `:ObsidianSearch` will attempt -
+	-- in that order - telescope.nvim, fzf-lua, and fzf.nvim, and use the
+	-- first one they find. By setting this option to your preferred
+	-- finder, you can attempt it first. Note that if the specified finder
+	-- is not installed, or if it the command does not support it, the
+	-- remaining finders will be attempted in the original order.
+	finder = "fzf-lua",
```
epwalsh, line 190: This is actually updated automatically from the README
```suggestion
```

README.md
```diff
@@ -149,6 +149,14 @@ return {
...
     -- Optional, set to true to force ':ObsidianOpen' to bring the app to the foreground.
     open_app_foreground = false,
+
+	-- Optional, by default commands like `:ObsidianSearch` will attempt -
+	-- in that order - telescope.nvim, fzf-lua, and fzf.nvim, and use the
+	-- first one they find. By setting this option to your preferred
+	-- finder, you can attempt it first. Note that if the specified finder
+	-- is not installed, or if it the command does not support it, the
+	-- remaining finders will be attempted in the original order.
+	finder = "fzf-lua",
```
epwalsh, line 159: The indentation looks off here
```suggestion

    -- Optional, by default commands like `:ObsidianSearch` will attempt -
    -- in that order - telescope.nvim, fzf-lua, and fzf.nvim, and use the
    -- first one they find. By setting this option to your preferred
    -- finder, you can attempt it first. Note that if the specified finder
    -- is not installed, or if it the command does not support it, the
    -- remaining finders will be attempted in the original order.
    finder = "fzf-lua",
```

lua/obsidian/command.lua
```diff
@@ -287,70 +295,73 @@ command.insert_template = function(client, data)
...
+                apply_template(template[1])
+              end)
+              return true
+            end,
+          }
+          require("telescope.builtin").find_files(opts)
+        end
+        choose_template()
+        return true
+      end
```
epwalsh, line 319: I think we should explicitly return false if `has_telescope` is false, and same for the others below. Just for clarity.

epwalsh (changes requested):
Hey @idanarye, I really like the approach you took. Only minor comments. 👍

lua/obsidian/command.lua
idanarye: Should I also verify in `run_first_supported` that the returned value is boolean?

idanarye: Another option is to throw a magic value from skipped implementations. That way `run_first_supported` can return the value of the implementation that succeeded.

epwalsh: > Another option is to throw a magic value from skipped implementations. That way run_first_supported can return the value of the implementation that succeeded.

I like that idea 👍

epwalsh (commented):
Thanks @idanarye, just a minor suggestion to make `util.run_first_supported` more readable.

lua/obsidian/util.lua
```diff
@@ -433,4 +433,72 @@ util.working_day_before = function(time)
...
+  for _, impl_name in ipairs(order) do
+    local impl_function = implementations[impl_name]
+    if impl_function then
+      local result = { pcall(impl_function) }
+      if result[1] then
+        return select(2, unpack(result))
+      elseif result[2] == IMPLEMENTATION_UNAVAILABLE then
+        table.insert(unavailable, impl_name)
+      else
+        error(result[2])
```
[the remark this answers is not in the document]
idanarye: But then we won't be able to use multiple returns from the implementation functions. `_run_with_finder_backend` doesn't use that (it doesn't even return a single value), but `run_first_supported` is a library function so I tried to make it as generic as possible.

Then again, maybe I'm just prematurely generalizing things...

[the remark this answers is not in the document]
epwalsh: Oooh, I see. Ok this is fine with me

epwalsh:
Looks like there's a styling and linting issue

idanarye:
@epwalsh There is one warning left, but it was not introduced by this PR (`main` has it: https://github.com/epwalsh/obsidian.nvim/actions/runs/4939421696/jobs/8830150709)

epwalsh:
Ah, sorry about that. Fixed.

epwalsh (approved):
LGTM!

PR #136, opened 2023-05-18
Respect disable_frontmatter when for daily notes

raindev:
Do not add frontmatter when disabled while using `:ObsidianToday` and `:ObsidianYesterday`.


epwalsh (approved):
Good catch, thank you!

PR #137, opened 2023-05-21
Fix bug in ObsidianTemplate with telescope

evtuhovich:
When i try to use :ObsidianTemplate i see such an error

Error detected while processing command line:
[Obsidian] ...ocal/share/nvim/lazy/obsidian.nvim/lua/obsidian/util.lua:465: ...hare/nvim/lazy/telescope.nvim/lua/telescope/mappings.lua:310: Attach mappings must always ret
urn a value. `true` means use default mappings, `false` means only use attached mappings


epwalsh (commented):
Thanks for the fix @evtuhovich! Can you just add an entry to the CHANGELOG?

epwalsh (approved):

PR #49, opened 2022-11-13
Provide custom format for dailynotes file name

AntoineBalaine:
Hello, 

I created an option to provide a custom date format to daily notes, in the config. 

I love every bit of obsidian.nvim but found that I had been using a different file format, so here it is. 
 
If you like this option, let me know and I can add a small mention in the readme. 

Cheers!

PR #138, opened 2023-05-30
Dailynotes custom format date

evtuhovich:
I've updated PR #49

Make daily_notes.date_format a string and now it works for both :ObsidianToday and :ObsidianYesterday


evtuhovich:
Make PR #49 works for ObsidianToday and ObsidianYesterday

epwalsh (approved):
Thanks @evtuhovich, I like this. Keeps it simpler.

PR #140, opened 2023-06-05
Fix running tests locally

epwalsh:
To run tests locally, do this:

```bash
make test PLENARY=~/.local/share/nvim/lazy/plenary.nvim/  # change the path here to where your plenary is installed
```


evtuhovich:
:heart: thanx @epwalsh

PR #142, opened 2023-06-09
fix: disable managed frontmatter for templates

### LINKED ISSUE #141: Don't manage frontmatter in templates

When the frontmatter of is managed by `obsidian.nvim` it is near impossible to edit template files. I usually writte these in markdown as well so `obsidian.nvim` treats them as notes inserting a frontmatter before saving the file.
I want the plugin to manage the frontmatter in most of my notes, but not in the templates because some of those aren't supposed to have a frontmatter.

I would suggest disabling the frontmatter in templates, by matching the path against the template path in the `BufWritePre` autocommand.

### Alternatives

Another more flexible solution might be allowing the user to optionally specify `disable_frontmatter` as a function. This can cover more cases where someone might want to opt out of the automatically managed frontmatter, but having access to `obsidian.nvim`'s internal normalization of `dir` might be beneficial for my use case.

### Additional context

The first option could be implemented by adding an early return to the callback in the `BufWritePre` autocommand or by extending the existing check.

```lua
if self.opts.templates ~= nil and vim.fn.matchstr(args.match, '^' .. tostring(self.dir / self.opts.templates.subdir / ".*")) ~= "" then
  return
end
```

This could possibly be extended by an additional option in the `opts.templates` table, to give users the ability to opt out if they want the frontmatter of templates to be managed.

Puerling:
closes #141

I decided to rebrand this as a fix, because it changes existing behavior.

If you want me to add the configuration option to reenable the current behavior (as described in #141) you need to decide what the default should be and how the two options for enabling the managed frontmatter should interact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Disabled managed frontmatter for files in the templates subdirectory.
 - A bug when `disable_frontmatter` is ignored for `ObsidianToday` and `ObsidianYesterday`.
 - A bug with `ObsidianTemplate` when using Telescope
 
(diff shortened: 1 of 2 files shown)


epwalsh (approved):
LGTM! Thank you @Puerling

PR #146, opened 2023-06-23
Add config to silent notifications and always create new notes in notes_subdir

TwIStOy:


epwalsh (commented):
Hey @TwIStOy, thanks for the PR. I have a couple comments:
1. It's not clear to me based on the name of the option what setting `never_current_dir` does.
2. There's a lot of places where we call `echo.info(..)`. Instead of wrapping all of those cases in an `if not opts.silent` block, maybe we could add the logic within `echo` module itself?

TwIStOy:
Thanks for your review!

1. Move the option from `config` to `config. completion` and rename it to `new_notes_location`. Maybe the name is more clear?
2. In the `echo` module, it's hard to get the instance of `obsidian.Client`. So I try to pass the `client` as a parameter into `echo.info`, I think it's a little wired. If there's some better options, please let me know. I will update the pr.

epwalsh:
Thanks @TwIStOy, I liked how you handled `new_notes_location`. For the `echo` module, I think it would make a little more sense to have a `log_level` setting instead of the `silent` option. And we could just pass `log_level` to `echo.(debug|info|warn|error)` instead of the `client` itself. How does that sound?

TwIStOy:
Sounds good. Change `silent` option to `log_level`.

lua/obsidian/config.lua
```diff
@@ -25,7 +26,9 @@ config.ClientOpts = {}
 config.ClientOpts.default = function()
   return {
     dir = vim.fs.normalize "./",
+    log_level = nil,
     notes_subdir = nil,
+    never_current_dir = false,
```
epwalsh, line 31:
```suggestion
```

epwalsh (approved):
Looks great!

Issue #53, opened 2022-11-23
Module 'obsidian' not found:


icholy:
### 🐛 Describe the bug

I'm attempting to install this plugin using packer:

``` lua
-- this is after nvim-cmp
use({
  "epwalsh/obsidian.nvim",
  after = "hrsh7th/nvim-cmp",
  setup = function()
    require("obsidian").setup({
      dir = "~/Documents/knowledge",
      completion = {
        nvim_cmp = true
      }
    })
  end
})
```

I'm getting this error:

```
packer.nvim: Error running setup for obsidian.nvim: [string "..."]:0: module 'obsidian' not found:
^Ino field package.preload['obsidian']
^Ino file './obsidian.lua'
^Ino file '/usr/share/luajit-2.1.0-beta3/obsidian.lua'
^Ino file '/usr/local/share/lua/5.1/obsidian.lua'
^Ino file '/usr/local/share/lua/5.1/obsidian/init.lua'
^Ino file '/usr/share/lua/5.1/obsidian.lua'
^Ino file '/usr/share/lua/5.1/obsidian/init.lua'
^Ino file '/home/icholy/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/obsidian.lua'
^Ino file '/home/icholy/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/obsidian/init.lua'
^Ino file '/home/icholy/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/rocks-5.1/obsidian.lua'
^Ino file '/home/icholy/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/rocks-5.1/obsidian/init.lua'
^Ino file './obsidian.so'
^Ino file '/usr/local/lib/lua/5.1/obsidian.so'
^Ino file '/usr/lib/x86_64-linux-gnu/lua/5.1/obsidian.so'
^Ino file '/usr/local/lib/lua/5.1/loadall.so'
```

### Versions

```
NVIM v0.9.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #150, opened 2023-07-13
docs(readme): adding packer.nvim setup

MichaBrugger:
Adding a `packer.nvim` base setup to the documentation, so it's easier to find. All credits and many thanks to @gplusplus314 for creating it, you saved me from a headache. For full thread/context, check #53.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,38 @@ return {
   end,
 }
 ```
+Basic setup using [´packer.vim´](https://github.com/wbthomason/packer.nvim):
+
+```lua
+  use({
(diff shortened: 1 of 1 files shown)


epwalsh (commented):
Thanks @MichaBrugger, this is great to have. One suggestion to keep things simple 👇

MichaBrugger:
@epwalsh, sure! Might make sense to list the options separate from the pkg manager at some point, but this should already give all package.nvim users an easy start 👍🏼

epwalsh:
Thanks, I agree with that. I'll follow up with another PR to reorganize the README

PR #151, opened 2023-07-13
Always treat paths relative to vault root

### LINKED ISSUE #147: Create new notes with path from vault directory

I use links within my templates with a path: [[Daily\ Notes/2023-06-06.md]] to create notes at a specific location when I follow the link (this is a "tomorrow" and "yesterday" link in my daily note). Currently this creates a note with the notes folder as the base; I think it should be the vault folder instead, to match the behaviour of Obsidian.

### Alternatives

_No response_

### Additional context

_No response_

epwalsh:
Fixes #147 (again). Simplifies the logic from #148.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed creating new notes when the title of the note contains a path.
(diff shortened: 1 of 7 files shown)


mscott99:
On links with a display bar [[path/file|display]], it throws an error, but still creates the file properly. Other than that it works great.

epwalsh:
Good catch! Fixed: 329098f

epwalsh:
We're lacking on unit tests here, but I've manually gone through all of the edge cases I can think of and it seems to be working as expected. I'm going to merge now b/c this is an important fix, but let me know if you find anymore issues. ✌️

PR #152, opened 2023-07-15
Fix ObsidianFollowLink with header links

epwalsh:
Fixes `:ObsidianFollowLink` when the link contains a header reference, e.g. `[[foo#Bar]]`.
Previously this would attempt to create a new note with the filename `foo#Bar.md`.

See https://github.com/epwalsh/obsidian.nvim/issues/50#issuecomment-1636676763


MohamedOsman1998:
dont mean to necro this but this doesnt work with links that have ^tags i.e. [[some_note^some_tag]] it still creates a note, seems very similar to this fix, just a different character, and would improve workflow by a lot. Thanks

epwalsh:
Hey @MohamedOsman1998, a PR would be appreciated. Thanks!

PR #154, opened 2023-07-21
Refactor `client.today` and `client.yesterday`

noriapi:
Since most parts of `client.today` and `client.yesterday` are the same code, i extracted it as `client._daily`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- (internal) Refactored daily note creation.
+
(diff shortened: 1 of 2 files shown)


noriapi:
Sorry, I don't know how the refactoring should be described in keep a changelog.

epwalsh:
Thanks @noriapi, this is great. For the CHANGELOG, I would just add something like this:

```markdown
### Changed

- (internal) Refactored daily note creation
```

noriapi:
> Thanks @noriapi, this is great. For the CHANGELOG, I would just add something like this:
> 
> ```
> ### Changed
> 
> - (internal) Refactored daily note creation
> ```

Thank you, I added it to CHANGELOG.

epwalsh (approved):
LGTM. Thanks @noriapi :)

Issue #128, opened 2023-04-24
Support for multiple vaults


sadtab:
### 🚀 The feature, motivation and pitch

The title says it all !

Great plug BTW, almost replaced the Obsidian app for me, I used the app just to visualize the graphs

epwalsh:
Thanks @sadtab! At the moment I think you could just call `obsidian.setup({...})` for each vault directory separately and it should work. Have you tried that, or did you have something else in mind?

sadtab:
I was thinking of something like [Neorg](https://github.com/nvim-neorg/neorg), You may have this plug already. What it does, is it can have multiple `workspaces` :
```lua
            config = {
                workspaces = {
                    work = "~/notes/work",
                    home = "~/notes/home",
                }
            }
```
Other than that, calling the `obsidian.setup({...})` with different paths seems like a decent workaround

aniketgm:
> At the moment I think you could just call `obsidian.setup({...})` for each vault directory separately and it should work.

I'm not sure how to do this. Do you mean changing the `dir` option each time when I want to switch to another vault ? A short example config would be helpful. I'm new to neovim. In the GUI Obsidian, I have two vaults "Work" and "Personal".

epwalsh:
Hey @aniketgm, I mean calling `obsidian.setup({})` twice in your `init.nvim` / `init.lua`. Like this for example:

```lua
obsidian.setup({
  dir = "~/notes/work",
  ...
})

obsidian.setup({
  dir = "~/notes/home",
  ...
})
```

vimcki:
i wanted to be able to select between 2 vaults, this telescope picker allows you to pick one of 2 vaults. It switches current directory and calls `require("obsidian").setup(obsopts)` with proper `obsopts.dir`. At the end there is auto group to call this picker on startup.
```lua
local pickers = require "telescope.pickers"
local finders = require "telescope.finders"
local conf = require("telescope.config").values
local actions = require "telescope.actions"
local action_state = require "telescope.actions.state"

obsopts = {
...
}

obsidian_picker = function(opts)
	opts = opts or {}
	pickers.new(opts, {
		prompt_title = "vaults",
		finder = finders.new_table {
			results = { "vault1", "vault2" }
		},
		sorter = conf.generic_sorter(opts),
		attach_mappings = function(prompt_bufnr, map)
			actions.select_default:replace(function()
				actions.close(prompt_bufnr)
				local selection = action_state.get_selected_entry()[1]
[...]

Issue #119, opened 2023-04-03
[FR] Set working directory to vault


daveriedstra:
### 🚀 The feature, motivation and pitch

Proposal: set the current working directory to the vault's directory when the plugin is loaded in a subdirectory of the vault.

My vault is organized into subdirectories. I often find myself having opened a note in a subdirectory directly, as opposed to navigating to the vault, opening nvim, and then opening the note. There are several nvim and plugin features which benefit from having a relevant working directory / CWD when interacting with notes across subdirectories:

- path completion (relevant to #75)
- path following (ie, `gf`'s default behaviour)
- search
- directory overview (eg, plugins like nerdtree)

Setting the working directory would allow users to take full advantage of these features.

Thanks for the fantastic plugin!

epwalsh:
Hey @daveriedstra, I'm happy to accept a PR for this. I'd like to keep it an optional feature, however it might make sense to make this the default behavior.

daveriedstra:
Thanks! I'll give it a shot if I can find the time.

metalelf0:
For this, I use [ahmedkhalf/project.nvim](https://github.com/ahmedkhalf/project.nvim). Here's my config (lazy.nvim format):

```lua
local M = {
	"ahmedkhalf/project.nvim",
	event = "VeryLazy",
}

M.config = function()
	local project = require("project_nvim")
	project.setup({
		active = true,
		on_config_done = nil,
		manual_mode = false,
		detection_methods = { "pattern" },
		patterns = { ".git", "_darcs", ".hg", ".bzr", ".svn", "Makefile", "package.json", ".project_root" },
		show_hidden = false,
		silent_chdir = true,
		ignore_lsp = {},
	})

	local telescope = require("telescope")
	telescope.load_extension("projects")
end

return M
```

This way, when I open a file and one of its parent directories contains a file that matches one of the `patterns`, the directory is changed.

epwalsh:
Unless I'm misunderstanding the feature request, this has been implemented via the `config.workspaces` and `config.detect_cwd` options.

Issue #60, opened 2022-11-27
Is it possible to change the working vault on the fly?


Paracelsus-Rose:
### 🚀 The feature, motivation and pitch

Looking for a way to change the working vault set up in config.  Not the best with lua or nvim, but this would be greatly used. Thinking of something along the lines of harpoon, where you can mark a few vaults, then open a modal and switch to each vault, then execute commands as needed.  If there is some way I can help, I would be willing to dive in and make a PR, just really not sure how to get started.  Thanks :-)

epwalsh:
Hey @Paracelsus-Rose, this is definitely doable. We could probably expose a new command that would change the Obsidian [client's directory](https://github.com/epwalsh/obsidian.nvim/blob/302cf774058f3a3f085aa03cd0ec7ca41d80cf21/lua/obsidian/init.lua#L28).

bassamsdata:
That's would be very helpful and a powerful command. 🔝

epwalsh:
We now have the `:ObsidianWorkspace` command.

PR #155, opened 2023-07-21
feat: support workspaces config option

weskoerber:
This PR adds the following functionality:

### "Workspaces" feature:
- A list of paths to directories
- Ability to switch between workspaces with `:ObsidianWorkspace`
  - This updates paths for `:ObsidianQuickSwitch`, `:ObsidianToday`, etc...
- If `detect_cwd` is true, attempt to resolve cwd in `workspaces`
  - If cwd does not exist in `workspaces`, open cwd as fallback vault
- If `detect_cwd` is false (default), use first workspace as working vault

### `:ObsidianWorkspace`:
- When called with no arguments, print the current workspace and its path
- When called with one argument, try to switch to that workspace

---

**Note:** `dir` is maintained for backward compatibility. If `dir` is specified, it's used as the default workspace (as long as `detect_cwd` is `false` [it is by default]).

To upgrade, move `dir` from the old config into the `workspaces` list. For example, if this is your current config:

```lua
obsidian.setup({
  dir = '~/Documents/notes',
})
```

Use this config:

```lua
obsidian.setup({
  workspaces = {
    {
      name = 'notes',
      path = '~/Documents/notes',
    }
  },
})
```

closes: #128

see also: #60, #119

### DIFF
--- a/doc/obsidian.txt
+++ b/doc/obsidian.txt
@@ -23,9 +23,10 @@ Obsidian app.
 - ⚙️ |obsidian-setup|
     - |obsidian-requirements|
     - |obsidian-install-and-configure|
+    - |obsidian-plugin-dependencies|
     - |obsidian-configuration-options|
-    - |obsidian-templates-support|
-    - |obsidian-using-nvim-treesitter|
+    - |obsidian-notes-on-configuration|
+    - |obsidian-using-templates|
 - 🐞 |obsidian-known-issues|
 - ➕ |obsidian-contributing|
 
@@ -34,28 +35,29 @@ Obsidian app.
 2. Features                                                *obsidian-features*
 
 - ▶️ Autocompletion for note references via nvim-cmp <https://github.com/hrsh7th/nvim-cmp> (triggered by typing `[[`)
-- 🏃 Go to a note buffer with `gf` when cursor is on a reference
+- 🏃 Optional passthrough for `gf` to enable Obsidian links without interfering with existing functionality
 - 💅 Additional markdown syntax highlighting and concealing for references
 
 
 COMMANDS                                                   *obsidian-commands*
 
-- `:ObsidianBacklinks` for getting a location list of references to the current buffer.
-- `:ObsidianToday` to create a new daily note.
-- `:ObsidianYesterday` to open (eventually creating) the daily note for the previous working day.
 - `:ObsidianOpen` to open a note in the Obsidian app.
     This command has one optional argument: the ID, path, or alias of the note to open. If not given, the note corresponding to the current buffer is opened.
 - `:ObsidianNew` to create a new note.
     This command has one optional argument: the title of the new note.
+- `:ObsidianQuickSwitch` to quickly switch to another notes in your vault, searching by its name using fzf.vim <https://github.com/junegunn/fzf.vim>, fzf-lua <https://github.com/ibhagwan/fzf-lua> or telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>.
+- `:ObsidianFollowLink` to follow a note reference under the cursor.
+- `:ObsidianBacklinks` for getting a location list of references to the current buffer.
+- `:ObsidianToday` to create a new daily note.
+- `:ObsidianYesterday` to open (eventually creating) the daily note for the previous working day.
+- `:ObsidianTemplate` to insert a template from the templates folder, selecting from a list using telescope.nvim <https://github.com/nvim-telescope/telescope.nvim> or one of the `fzf` alternatives.
+    See |obsidian-"using-templates"| for more information.
 - `:ObsidianSearch` to search for notes in your vault using ripgrep <https://github.com/BurntSushi/ripgrep> with fzf.vim <https://github.com/junegunn/fzf.vim>, fzf-lua <https://github.com/ibhagwan/fzf-lua> or telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>.
     This command has one optional argument: a search query to start with.
-- `:ObsidianQuickSwitch` to quickly switch to another notes in your vault, searching by its name using fzf.vim <https://github.com/junegunn/fzf.vim>, fzf-lua <https://github.com/ibhagwan/fzf-lua> or telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>.
 - `:ObsidianLink` to link an in-line visual selection of text to a note.
     This command has one optional argument: the ID, path, or alias of the note to link to. If not given, the selected text will be used to find the note with a matching ID, path, or alias.
 - `:ObsidianLinkNew` to create a new note and link it to an in-line visual selection of text.
     This command has one optional argument: the title of the new note. If not given, the selected text will be used as the title.
-- `:ObsidianFollowLink` to follow a note reference under the cursor.
-- `:ObsidianTemplate` to insert a template from the templates folder, selecting from a list using telescope.nvim <https://github.com/nvim-telescope/telescope.nvim> or one of the `fzf` alternatives.
 
 
 DEMO                                                           *obsidian-demo*
@@ -77,14 +79,15 @@ REQUIREMENTS                                           *obsidian-requirements*
 Search functionality (e.g. via the `:ObsidianSearch` and
 `:ObsidianQuickSwitch` commands) also requires telescope.nvim
 <https://github.com/nvim-telescope/telescope.nvim> or one of the `fzf`
-alternatives (see below).
+alternatives (see |obsidian-plugin-dependencies| below).
 
 
 INSTALL AND CONFIGURE                         *obsidian-install-and-configure*
 
-To configure Obsidian.nvim you just need to call `require("obsidian").setup({
+To configure obsidian.nvim you just need to call `require("obsidian").setup({
 ... })` with the desired options. Here are some examples using different plugin
-managers. The full set of configuration options are listed |obsidian-below|.
+managers. The full set of |obsidian-plugin-dependencies| and
+|obsidian-configuration-options| are listed below.
 
 
 USING LAZY.NVIM ~
@@ -100,41 +103,13 @@ USING LAZY.NVIM ~
         -- Required.
         "nvim-lua/plenary.nvim",
     
-        -- Optional, for completion.
-        "hrsh7th/nvim-cmp",
-    
-        -- Optional, for search and quick-switch functionality.
-        "nvim-telescope/telescope.nvim",
-    
-        -- Optional, an alternative to telescope for search and quick-switch functionality.
-        -- "ibhagwan/fzf-lua"
-    
-        -- Optional, another alternative to telescope for search and quick-switch functionality.
-        -- "junegunn/fzf",
-        -- "junegunn/fzf.vim"
-    
-        -- Optional, alternative to nvim-treesitter for syntax highlighting.
-        "godlygeek/tabular",
-        "preservim/vim-markdown",
+        -- see below for full list of optional dependencies 👇
       },
       opts = {
         dir = "~/my-vault",  -- no need to call 'vim.fn.expand' here
     
         -- see below for full list of options 👇
       },
-      config = function(_, opts)
-        require("obsidian").setup(opts)
-    
-        -- Optional, override the 'gf' keymap to utilize Obsidian's search functionality.
-        -- see also: 'follow_url_func' config option below.
-        vim.keymap.set("n", "gf", function()
-          if require("obsidian").util.cursor_on_markdown_link() then
-            return "<cmd>ObsidianFollowLink<CR>"
-          else
-            return "gf"
-          end
-        end, { noremap = false, expr = true })
-      end,
     }
 <
 
@@ -144,6 +119,12 @@ USING PACKER.NVIM ~
 >lua
     use({
       "epwalsh/obsidian.nvim",
+      requires = {
+        -- Required.
+        "nvim-lua/plenary.nvim",
+    
+        -- see below for full list of optional dependencies 👇
+      },
       config = function()
         require("obsidian").setup({
           dir = "~/my-vault",
@@ -153,6 +134,22 @@ USING PACKER.NVIM ~
       end,
     })
 <
+
+
+PLUGIN DEPENDENCIES                             *obsidian-plugin-dependencies*
+
+The only required plugin dependency is plenary.nvim
+<https://github.com/nvim-lua/plenary.nvim>, but there are a number of optional
+dependencies that enhance the obsidian.nvim experience:
+
+- hrsh7th/nvim-cmp <https://github.com/hrsh7th/nvim-cmp>: for completion of note references.
+- nvim-telescope/telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>: for search and quick-switch functionality.
+- ibhagwan/fzf-lua <https://github.com/ibhagwan/fzf-lua>: an alternative to telescope for search and quick-switch functionality.
+- junegunn/fzf <https://github.com/junegunn/fzf> and junegunn/fzf.vim <https://github.com/junegunn/fzf.vim>: another alternative to telescope for search and quick-switch functionality.
+- godlygeek/tabular <https://github.com/godlygeek/tabular> and preservim/vim-markdown <https://github.com/preservim/vim-markdown>: alternative to nvim-treesitter for syntax highlighting.
+
+If you choose to use any of these you should include them in the "dependencies"
+or "requires" field of the obsidian.nvim plugin spec for your package manager.
 
 
 CONFIGURATION OPTIONS                         *obsidian-configuration-options*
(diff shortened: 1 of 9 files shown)


epwalsh (commented):
FYI you don't need to update `doc/obsidian.txt`. That file is generated automatically from the README. So just make those same changes in the README. I recently added a [CONTRIBUTING](https://github.com/epwalsh/obsidian.nvim/blob/main/.github/CONTRIBUTING.md) guide that goes over that and some other tips for getting CI to pass.

weskoerber:
I added the `workspaces` field and kept the `dir` field.

The idea is that the `workspaces` is assigned a value by the user when calling `obsidian.setup`. For example:

```lua
require('obsidian').setup({
  workspaces = { 
    '~/vault-1',
    '~/vault-2',
  },
}
```

Then, we loop through the `workspaces` table and compare each entry to the current working directory. The first matching value is assigned to `dir`.

One problem with this implementation is that the behavior is not consistent depending on whether the user passes in `dir` or not:
- If `dir` is not assigned a value by the user and `workspaces` contains the current working directory, `dir` will be assigned the  value of the current working directory from `workspaces`
- If `dir` is assigned a value by the user and `workspaces` contains the current working directory, the value assigned to `dir` by the user will be reassigned (by `client.ConfigOpts.normalize()`) the value of the current working directory from `workspaces`
- If `dir` is assigned a value by the user and `workspaces` does not contain the current working directory, `dir` keeps its assignment
- If `dir` is not assigned a value by the user, and `workspaces` does not contain the current working directory, `dir` defaults to the current working directory in `obsidian.new()`

A potential solution to this inconsistency is to ignore what the user assigns to `dir`.

However, the last bullet point has me questioning this feature altogether. If `dir` defaults to the working directory, what's the point of workspaces? Wouldn't falling back to the working directory effectively make any directory a valid workspace?

weskoerber:
> FYI you don't need to update `doc/obsidian.txt`. That file is generated automatically from the README. So just make those same changes in the README. I recently added a [CONTRIBUTING](https://github.com/epwalsh/obsidian.nvim/blob/main/.github/CONTRIBUTING.md) guide that goes over that and some other tips for getting CI to pass.

Ahh, I should have read that first...

epwalsh:
Hey @weskoerber, I think `dir` and `workspaces` should be mutually exclusive.

> If `dir` is not assigned a value by the user and `workspaces` contains the current working directory, `dir` will be assigned the value of the current working directory from `workspaces`.

In this case I `dir` should be assigned the corresponding path from `workspaces`, not necessary the current working directory, which could be a subdirectory of the path from `workspaces`.

weskoerber:
> In this case I dir should be assigned the corresponding path from workspaces, not necessary the current working directory, which could be a subdirectory of the path from workspaces.

Using the following config...

```lua
obsidian.setup({
  workspaces = {
    '~/notes',
  },
})
```

...and assuming the working directory is `~/notes/personal`, `dir` should be assigned `~/notes` instead of `~/notes/personal`?

What if `workspaces` is empty, or `workspaces` does not contain an acceptable value for the current working directory?

> Hey @weskoerber, I think dir and workspaces should be mutually exclusive.

My initial idea is that `dir` and `workspaces` was not mutually exclusive; if `vim.fn.getcwd()` matched a value in `workspaces`, then we'd assign the value to `dir`, using it as our current vault. Basically, `workspaces` is a list of vaults, and only one can be "active" at a time.

Is that different than your expectation? How do you expect the workspaces feature to work?

weskoerber:
I went back to #128 and looked at [neorg](https://github.com/nvim-neorg/neorg) workspaces. They say workspaces are "basically isolated directories that you can jump between".

My initial hack I wrote up in #128 was just to support opening a directory (e.g. `cd ~/notes/personal && nvim`) and using that as a vault, essentially setting the `dir` config option dynamically.

However, I think @sadtab's idea of workspaces were more similar to [neorg's](https://github.com/nvim-neorg/neorg) workspaces, and my little hack misses that mark.

If we can jump between workspaces (e.g. `:ObsidianSwitchWorkspace work`), then what's the purpose of `dir`?

sadtab:
I think `workspace`, `dir` and `vault` should be all the same. Let me elaborate the example of neorg. In neorg we have the concept of `workspaces`, each workspace is just a path to a folder. One of them is always set as default, whether the user sets it explicitly or it will just pick one of them.   
Any command of Neorg (like new note, search notes etc) is base on the `CURRENT` workspace, there is even a command to print the current workspace. If the user changes the workspace (`:Neorg workspace myAwesomeWorkspace` for example)  every command (like new note, search notes etc) will be applied to the current workspace.

Based on the current implementation of obsidian.nvim, I think the lease intrusive approach can be the same. Let the user to define a list of vaults and pick one as default, there is always one vault set as the _current vault_. 

Don't bother about the cwd or try to detect the cwd. Because its complex, bug-prone, and also inconvenient in case you are working on your source codes and you want just to take a note quickly. The user knows its default vault, if they want to choose another vault, it changes the vault and then takes the note.

weskoerber:
Thanks for the feedback guys!

@epwalsh:

> Ok so I think there's really two different features here:
>
> 1. Supporting more than one "static" vault directory.
> 2. Dynamically changing the vault dir to cwd.

I agree, although I think your first point can be split into 2 parts

1a. #128 - support for multiple vaults
1b. #60 - change vault on the fly
2. #119 - use working directory as vault

However, I don't think I understand your second point completely (also #119), because I don't know that it's necessary with 1a and 1b. But I can see the value in changing the vault dir to cwd if the vault dir isn't set with 1a and 1b. Lemme know what you think.

---

@sadtab:

> I think `workspace`, `dir` and `vault` should be all the same.

This was my initial assumption as well. I thought that `workspaces` would essentially be a list of paths to obsidian vaults, and `dir` would be the currently selected workspace.

> One of them is always set as default, whether the user sets it explicitly or it will just pick the first one

IMHO, a more useful implementation would determine if the current directory exists in the `workspaces` list. If it does, we use it. However, I don't really know the best action to take if it doesn't exist. Maybe then we can fall back to a default workspace, or pick the first one like you said.

> Don't bother about the cwd or try to detect the cwd. Because it's complex, bug-prone, and also inconvenient in the very common case where you are working on your source codes and you want just to take a note quickly.

The user also knows the `cwd` when neovim is opened, so one could assume that if neovim is opened in a directory that exists in the `workspaces` list, the user intended to open that vault, rather than the default vault.

Let me know what y'all think!

weskoerber:
I misunderstood #119. The author of that issue said:

> set the current working directory to the vault's directory when the plugin is loaded in a subdirectory of the vault

Which is different than your second point.

epwalsh:
Thanks for the continued effort on this @weskoerber :) I'm very busy with my real job this week so I might not be able to give another review until next week.

weskoerber:
No worries, I'm in no rush! I got busy a couple weeks ago too, I know how it goes...

okuuva:
Really looking forward to this! I'm just in the process of setting up Obsidian and the ability to separate work and personal notes is a must have for me..

okuuva:
>- If default_workspace is set in the config, open that workspace
>- If default_workspace is not set in the config, and workspaces list contains the cwd, open workspace that corresponds to cwd
>- If default_workspace is not set, and workspaces list does not contain the cwd, the cwd is used as a fallback

This whole `default_workspace` business seems overly complicated to me. I just don't see what a separate setting brings to the table over selecting the first workspace from the list as a default. I would assume it's a more useful default over `cwd` anyway: that way you would have access to your default vault regardless where you happened to open nvim. To me, having to set another config option to achieve that behavior seems like a hassle and having it to default to `cwd` seems confusing. But maybe that's just me.

weskoerber:
> that way you would have access to your default vault regardless where you happened to open nvim

Yep, that's exactly the intended behavior. It's actually not new functionality, just renamed from `dir` because I thought the name was a bit more clear, although this breaking change is totally unnecessary.

> I just don't see what a separate setting brings to the table over selecting the first workspace from the list as a default

The intention of `default_workspace` was to allow the user to specify a specific vault to open at startup, even if the user opened neovim in a directory that exists in `workspaces`. If `default_workspace` wasn't set and we instead used the first item in `workspaces`, but neovim was opened in a directory that exists in `workspaces`, then that workspace (essentially cwd) would be opened.

> having to set another config option to achieve that behavior seems like a hassle and having it to default to `cwd` seems confusing

Admittedly, `default_workspace` is dual-purpose: decide whether we should detect `cwd`, and set a default vault. Maybe a better approach could add a setting like `detect_cwd` that controls whether we detect `cwd` or not: 
- if true, we detect if `cwd` is in `workspaces` and use it as our working vault, otherwise we fall back to the first workspace in `workspaces`
- if false, we fall back to the first workspace in `workspaces`

okuuva:
> > having to set another config option to achieve that behavior seems like a hassle and having it to default to `cwd` seems confusing
> 
> Admittedly, `default_workspace` is dual-purpose: decide whether we should detect `cwd`, and set a default vault. Maybe a better approach could add a setting like `detect_cwd` that controls whether we detect `cwd` or not:
> 
>     * if true, we detect if `cwd` is in `workspaces` and use it as our working vault, otherwise we fall back to the first workspace in `workspaces`
> 
>     * if false, we fall back to the first workspace in `workspaces`

Yeah, the dual-purpose was what made it a bit confusing. A boolean for tracking/detecting/following `cwd` and the logic you described sounds like a much more intuitive approach.

weskoerber:
Lua hash maps are unordered so the order of `workspaces` may change. This means that the first item in `workspaces` may differ between consecutive invocations of `obsidian.setup()`. In order to implement this, I refactored `workspaces`; instead of using the name of the workspace as the key, I moved the name into the workspace object. `workspaces` is now a "array" (a table indexed with integers, rather than strings). The order of `workspaces` will now be maintained.

okuuva:
> Lua hash maps are unordered so the order of `workspaces` may change. This means that the first item in `workspaces` may differ between consecutive invocations of `obsidian.setup()`. In order to implement this, I refactored `workspaces`; instead of using the name of the workspace as the key, I moved the name into the nested table. `workspaces` is now a "array" (a table indexed with integers, rather than strings). The order of `workspaces` will now be maintained.
> 
> This makes the definition of workspaces a little more verbose, but it makes the implementation cleaner IMO:
> 
> ```lua
> obsidian.setup({
>   workspaces = {
>     {
>       name = 'personal',
>       path = '~/notes/personal',
>     },
>     {
>       name = 'work',
>       path = '~/notes/work',
>     }
>   },
>   detect_cwd = false, -- default
> })
> ```

If we want named workspaces then this is the way to go. But is the naming really necessary? Just wondering the use case here and can't figure out anything but easier quick switch using a command. If using a picker like telescope then the name doesn't really matter much.

But the logic is now much easier to follow and I agree, the implementation looks much clearer as well. Kudos!

lua/obsidian/command.lua
```diff
@@ -647,6 +647,33 @@ command.check_health = function(client, _)
...
+
+  if not workspace then
+    echo.err("Workspace '" .. data.args .. "' does not exist", client.opts.log_level)
+    return
+  end
+
+  client.current_workspace = workspace
+
+  echo.info("Switching to workspace '" .. workspace.name .. "' (" .. workspace.path .. ")", client.opts.log_level)
+  client.dir = Path:new(workspace.path)
```
epwalsh, line 674: Same suggestion here about normalizing:

```suggestion
  client.dir = Path:new(vim.fs.normalize(tostring(workspace.path)))
```

lua/obsidian/workspace.lua
```diff
@@ -0,0 +1,71 @@
...
+end
+
+---Determines if cwd is a workspace
+---
+---@param workspaces table<obsidian.Workspace>
+---@return obsidian.Workspace|nil
+workspace.get_workspace_from_cwd = function(workspaces)
+  local cwd = vim.fn.getcwd()
+  local _, value = next(vim.tbl_filter(function(w)
+    if w.path == cwd then
```
epwalsh, line 33: We might need to normalize these paths before comparing them, otherwise it seems a bit brittle.

weskoerber:
> Thank you all for the discussion! I just have a few minor requests and then I think we can get this merged ✅

Sorry for the delay. I'll have some time this weekend to address these concerns!

MathieuDR:
I would love this feature, what's the state of this? @weskoerber

I would like to thank you already for your work!

john-okeefe:
Is there any update on this? I'm really excited for this feature!

sebassdc:
Looking forward! great work on this!

lua/obsidian/init.lua
```diff
@@ -24,7 +26,9 @@ local client = {}
 ---@return obsidian.Client
 obsidian.new = function(opts)
   local self = setmetatable({}, { __index = client })
-  self.dir = Path:new(vim.fs.normalize(tostring(opts.dir and opts.dir or "./")))
+
+  self.current_workspace = obsidian.workspace.get_from_opts(opts)
+  self.dir = Path:new(self.current_workspace.path)
```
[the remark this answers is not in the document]
weskoerber: Hi! I finally have time to swing back around on these...

The workspace paths should already be normalized.

Here, in [`workspace.new()`](https://github.com/weskoerber/obsidian.nvim/blob/46b4a83a1ff965f4fe479f3d3884889a4ea82a4f/lua/obsidian/workspace.lua#L17), the path is normalized:

```lua
workspace.new = function(name, path)
  local self = setmetatable({}, { __index = workspace })

  self.name = name
  self.path = vim.fs.normalize(path)

  return self
end
```

Paths are also normalized when reading from the config in [`client.ClientOpts.normalize()`](https://github.com/weskoerber/obsidian.nvim/blob/46b4a83a1ff965f4fe479f3d3884889a4ea82a4f/lua/obsidian/config.lua#L76):

```lua
  for key, value in pairs(opts.workspaces) do
    opts.workspaces[key].path = vim.fs.normalize(tostring(value.path))
  end
```

There could be an edge-case I haven't fully considered, but I think it's safe to omit the redundant path normalization. However, if you still prefer normalizing the paths where you suggested I won't argue with you :smile:

lua/obsidian/command.lua
weskoerber: See [my comment](https://github.com/epwalsh/obsidian.nvim/pull/155#discussion_r1361333890) in a previous suggestion

lua/obsidian/workspace.lua
weskoerber: See [my comment](https://github.com/epwalsh/obsidian.nvim/pull/155#discussion_r1361333890) in a previous suggestion

epwalsh (commented):
Thanks for getting back to this @weskoerber! Just a couple more comments

doc/obsidian.txt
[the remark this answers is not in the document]
weskoerber: Ugh sorry, I might have resolved conflicts in this file poorly when I rebased.

lua/obsidian/config.lua
```diff
@@ -1,11 +1,14 @@
...
+local workspace = require "obsidian.workspace"
 
 local config = {}
 
 ---[[ Options specs ]]---
 
 ---@class obsidian.config.ClientOpts
----@field dir string
+---@field dir string|?
+---@field workspaces table
```
epwalsh, line 10: I think this has the `obisidan.Workspace` type, right?
```suggestion
---@field workspaces obsidian.Workspace
```

weskoerber: It's a table of `obsidian.workspace`s:

```lua
require('obsidian').setup({
    -- ...

    workspaces = {
        {
            name = 'personal',
            path = '~/Documents/notes/personal',
        },
        {
            name = 'work',
            path = '~/Documents/notes/acsd',
        },
    }
    -- ...
)}
```

lua/obsidian/config.lua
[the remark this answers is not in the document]
weskoerber: Fixed in f422f98

lua/obsidian/init.lua
```diff
@@ -37,7 +41,7 @@ end
 ---@return obsidian.Client
 obsidian.new_from_dir = function(dir)
   local opts = config.ClientOpts.default()
-  opts.dir = vim.fs.normalize(dir)
+  opts.workspaces = vim.tbl_extend("force", { obsidian.workspace.new("test_vault", dir) }, opts.workspaces)
```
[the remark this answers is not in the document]
weskoerber: Fixed with 1e1ba52

lua/obsidian/init.lua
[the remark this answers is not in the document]
weskoerber: Fixed with 598bc0a

lua/obsidian/command.lua
weskoerber: Fixed with 598bc0a

doc/obsidian.txt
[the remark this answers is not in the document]
weskoerber: Fixed with 48ae1f1

lua/obsidian/config.lua
epwalsh: Ah, right, then we can give it a type hint like this
```suggestion
---@field workspaces obsidian.Workspace[]|?
```

epwalsh (approved):
LGTM! Thanks for getting this over the finish line @weskoerber :)

PR #156, opened 2023-07-22
Small documentation changes

shakesbeare:
In accordance with some new user issues which can be found [here](https://youtu.be/SkMQStnqCZ0?t=939), I'm proposing a few small changes to the docs in addition to some other issues I'm going to file as well. 

The two things going on here are:
1) Adjusted the language in the Features section to make it clear that `gf` passthrough is not enabled by default. 
2) Reordered the list of commands to try and put the very basic functionality you might want to initially test closer to the top.


epwalsh (approved):
Thanks @shakesbeare, LGTM

PR #160, opened 2023-07-24
Remove leading | if note_id_func returns '' or nil

### LINKED ISSUE #159: Ability to remove id from note name

Currently when creating a new note we are forced to prepend an ID.
image
I would like the ability to remove the need for the prepended ID, and middle `|` so it's just the page name which looks cleaner in the link.
The closest I can get is to remove the leading id with the following config
```lua
note_id_func = function (_)
    return ''
end
```
However that still has the leading `|` character.
image
I think something simple as checking if the id function returns nothing don't prepend the `|` char would solve this problem.

### Alternatives

I haven't considered alternatives as it seems pretty straightforward and minimally impacting to add.

### Additional context

_No response_

NachoxMacho:
Resolves #159
Adjusts `cmp_obsidian_new` so that if the `note_id_func` function returns an empty string or nil, there is no longer a leading `|`. 
I've confirmed on my end that this works with the following config:
```lua
    note_id_func = function (_)
        return ''
    end
```
image

I believe I've updated all relevant documentation and changelogs in accordance with the contributing guidelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,8 +169,7 @@ This is a complete list of all of the options that can be passed to `require("ob
     prepend_note_id = true
   },
 
-  -- Optional, customize how names/IDs for new notes are created. If nil or an empty string
-  -- is returned the id will be ignored.
+  -- Optional, customize how names/IDs for new notes are created.
   note_id_func = function(title)
     -- Create note IDs in a Zettelkasten format with a timestamp and a suffix.
     -- In this case a note with the title 'My new note' will given an ID that looks
--- a/README.md
+++ b/README.md
@@ -165,7 +165,8 @@ This is a complete list of all of the options that can be passed to `require("ob
     --  * "notes_subdir" - put new notes in the default notes subdirectory.
     new_notes_location = "current_dir",
 
-    -- Whether to add the output of the node_id_func to new notes in autocompletion
+    -- Whether to add the output of the node_id_func to new notes in autocompletion.
+    -- E.g. "[[Foo" completes to "[[foo|Foo]]" assuming "foo" is the ID of the note.
     prepend_note_id = true
   },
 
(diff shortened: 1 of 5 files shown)


epwalsh (changes requested):
Hey @NachoxMacho, I'm nervous about allowing `note_id_func` to return an empty string or nil because that departs from the original API and might have unintended consequences elsewhere. Alternatively I'd prefer adding another configuration option to turn this behavior on.

NachoxMacho:
I understand there could be unintended consequences of this change, I have only tested it briefly to my use cases. So something like a new top level boolean config option like `prepend_note_id` would be more desired here? If that's false then the code previously modified would just not prepend the `note_id_func` output, and it defaults to true? Is that more along what you had in mind to accomplish this?

epwalsh:
Yes that sounds great. Except instead of a top-level config option, let's make it part of the `completion` section. Sound good?

NachoxMacho:
Sounds good to me, I'll make the adjustments to this PR later today

NachoxMacho:
@epwalsh I've updated the code accordingly and validated on my system that toggling the flag adds the listed effect. The default for the new field is `true`so it doesn't break current configurations.

README.md
```diff
@@ -163,7 +163,10 @@ This is a complete list of all of the options that can be passed to `require("ob
     -- Where to put new notes created from completion. Valid options are
     --  * "current_dir" - put new notes in same directory as the current buffer.
     --  * "notes_subdir" - put new notes in the default notes subdirectory.
-    new_notes_location = "current_dir"
+    new_notes_location = "current_dir",
+
+    -- Whether to add the output of the node_id_func to new notes in autocompletion
```
epwalsh, line 168:
```suggestion
    -- Whether to add the output of the node_id_func to new notes in autocompletion.
    -- E.g. "[[Foo" completes to "[[foo|Foo]]" assuming "foo" is the ID of the note.
```

epwalsh (approved):
LGTM! Thanks @NachoxMacho

PR #161, opened 2023-07-26
Automatically configure 'gf' passthrough

### LINKED ISSUE #157: Make `gf` keybinding a part of the config

Making `gf` pass through the default behavior seems like it would enhance the new-user experience greatly. I'm proposing we move the keybind configuration into the plugin itself and have a flag in the config like "configure_gf_passthrough" or "manage_keybinds" (in case there are other keybindings that might be added in the future). 

### Alternatives

The one question I have about this is whether or not to turn it on by default. It seems like having plugins manage their own keybinds is not an unheard of thing, but it might be frustrating if our implementations for keybinds end up having any errors that result in unexpected breakages even while not in the vault. For example, messing with `gf` could cause issues with someone just trying to use the command normally. 

Personally, I think that since the `gf` keybind shouldn't affect anything if the user isn't on a markdown link, it probably is fine to turn it on by default. But we should probably add a check whether or not the open buffer is in the vault directory at the same time to ensure there aren't any issues. 

On the other hand, turning it on by default may cause existing configurations to break if they have a custom bind already setup for `gf`. Plus, if there are ever any other keybinds added, they might be more disruptive and may want to be turned off by default. 

### Additional context

_No response_

epwalsh:
Fixes #157.

- Adds a `mappings` field to the config.
- obsidian.nvim will now automatically configure the 'gf' passthrough unless this keybinding was already set or you override the `mappings` field in your obsidian.nvim config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,19 +86,6 @@ return {
 
     -- see below for full list of options 👇
   },
-  config = function(_, opts)
-    require("obsidian").setup(opts)
-
-    -- Optional, override the 'gf' keymap to utilize obsidian.nvim's search functionality.
-    -- see also: 'follow_url_func' config option below.
-    vim.keymap.set("n", "gf", function()
-      if require("obsidian").util.cursor_on_markdown_link() then
-        return "<cmd>ObsidianFollowLink<CR>"
-      else
-        return "gf"
-      end
-    end, { noremap = false, expr = true })
-  end,
 }
 ```
 
@@ -172,6 +159,12 @@ This is a complete list of all of the options that can be passed to `require("ob
     -- Whether to add the output of the node_id_func to new notes in autocompletion.
     -- E.g. "[[Foo" completes to "[[foo|Foo]]" assuming "foo" is the ID of the note.
     prepend_note_id = true
+  },
+
+  -- Optional, keybindings.
+  keybindings = {
+    -- Overrides the 'gf' keybinding to work on markdown/wiki links within your vault.
+    gf_passthrough = true,
   },
 
   -- Optional, customize how names/IDs for new notes are created.
--- a/README.md
+++ b/README.md
@@ -239,6 +239,17 @@ This is a complete list of all of the options that can be passed to `require("ob
 - obsidian.nvim will set itself up as an nvim-cmp source automatically when you enter a markdown buffer within your vault directory, you do **not** need to specify this plugin as a cmp source manually.
 - If you use `vim-markdown` you'll probably want to disable its frontmatter syntax highlighting (`vim.g.vim_markdown_frontmatter = 1`) which I've found doesn't work very well.
 - The `notes_subdir` and `note_id_func` options are not mutually exclusive. You can use them both. For example, using a combination of both of the above settings, a new note called "My new note" will assigned a path like `notes/1657296016-my-new-note.md`.
+- If you want the `gf` passthrough functionality but you've already overridden the `gf` keybinding, just change your `gf` mapping definition to something like this:
+
+  ```lua
+  vim.keymap.set("n", "gf", function()
+    if require("obsidian").util.cursor_on_markdown_link() then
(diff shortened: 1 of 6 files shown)


test/obsidian/note_spec.lua
```diff
@@ -20,7 +20,7 @@ describe("Note", function()
     local note = Note.from_file "README.md"
     assert.equals(note.id, "README")
     assert.equals(#note.aliases, 1)
-    assert.equals(note.aliases[1], "Obsidian.nvim")
+    assert.equals(note.aliases[1], "obsidian.nvim")
```
shakesbeare, line 23: This seems like it might be unrelated and might be deserving of its own message in the commit log

README.md
```diff
@@ -86,19 +86,6 @@ return {
 
     -- see below for full list of options 👇
   },
-  config = function(_, opts)
-    require("obsidian").setup(opts)
-
-    -- Optional, override the 'gf' keymap to utilize obsidian.nvim's search functionality.
-    -- see also: 'follow_url_func' config option below.
-    vim.keymap.set("n", "gf", function()
```
shakesbeare, line 94: In the spirit of updating the docs, we might want to leave this information somewhere in the README in to provide some direction for users who wish to manage their commands manually. For example, I don't think we document anywhere that this helper function `cursor_on_markdown_link()` exists, which might be very helpful information for power users.

lua/obsidian/init.lua
```diff
@@ -91,6 +91,24 @@ obsidian.setup = function(opts)
...
     end
+
+    -- Keybindings...
+    -- 'gf' passthrough
+    if opts.keybindings.gf_passthrough then
+      local gf_set_by = vim.fn.mapcheck("gf", "n")
+      if gf_set_by == "" then
+        -- 'gf' has not been overridden by user or other plugin so we can safely set it.
+        vim.keymap.set("n", "gf", function()
+          if require("obsidian").util.cursor_on_markdown_link() then
```
shakesbeare, line 102: I think there should be an additional check here to ensure that the current buffer is located within the obsidian vault. Could cause problems if someone happens to be writing links formatted like Obsidian links elsewhere. 

Maybe it should be an optional parameter for `gf_passthrough`? `only_allow_in_vault` or something like that.

README.md
```diff
@@ -174,6 +161,12 @@ This is a complete list of all of the options that can be passed to `require("ob
     prepend_note_id = true
   },
 
+  -- Optional, keybindings.
+  keybindings = {
+    -- Overrides the 'gf' keybinding to work on markdown/wiki links within your vault.
+    gf_passthrough = true,
```
shakesbeare, line 167: This is exactly what I described in my issue suggesting this, but I'm now wondering if a more robust system might be preferable? For instance, like what can be found in [here](https://github.com/hrsh7th/nvim-cmp). 

Could potentially come up with a system like that for mappings to many of our commands with options (like passthrough) to modify their behavior from default. 

One the one hand, it seems a bit more transparent in case a user might want different keybind(s) than we set in the plugin, but it also might be a waste of effort if this is the only bind we ever set up.

shakesbeare (commented):
Just a few comments

test/obsidian/note_spec.lua
epwalsh: This was just because I changed the title of the README from "Obsidian.nvim" to "obsidian.nvim" to be consistent with the repo name.

README.md
epwalsh: Yeup, agreed: 3da669d

lua/obsidian/init.lua
epwalsh: Since we create this keymap within the lazy setup function and we have `buffer = true` in the keymap options, it's already only going to be set for buffers in the vault.

README.md
epwalsh: Actually I like that a lot. 53ea312

PR #164, opened 2023-08-03
fix bug in template check if there's magic chars in path

TwIStOy:


epwalsh (commented):
Thanks @TwIStOy! Looking at Lua patterns gives me a headache 😅

Could you just move the "escape special characters" logic to a util function (`lua/obsidian/util.lua`) and add some test cases to `test/obsidian/util_spec.lua` for it?

epwalsh (approved):
Beautiful! Thank you!

PR #166, opened 2023-08-03
Commands for opening notes in a vertical split

### LINKED ISSUE #165: [Feature] Open notes in a vertical split

I typically write notes at the same time as working, and it would be nice to be able to open the notes in a vertical split to record my notes side by side with my work.

I see this as being a non-breaking change, as it would add distinctly new commands to the plugin without affecting current features.

### Alternatives

_No response_

### Additional context

_No response_

stefanlogue:
Added some commands to open notes in a vertical split:
- ObsidianTodayVertical
- ObsidianYesterdayVertical
- ObsidianNewVertical

Also updated README.md to reflect changes.

Closes #165


epwalsh:
Hey @stefanlogue, thanks for this. I would like to keep the overall number of commands to a minimum though, so IMO a better option would be to add an argument to the existing commands which determines how to open those buffers (vertical, horizontal, or replace current buffer). What do you think?

stefanlogue:
Yeah that sounds like a better implementation, I can have a look at implementing that tomorrow/early next week. Should I close this PR and open a fresh one, as technically it's a different feature set?

epwalsh:
@stefanlogue up to you! Fine with me if you just make those changes in this PR.

stefanlogue:
I think this could become quite tricky in the case of new notes, which already have an optional arg for the name of the note. If a user doesn't want to name their note, but _does_ want to open it in a split, how do we discern that the passed argument is supposed to target the opening strategy rather than the name?

Your input on this would be much appreciated, as I want to avoid breaking any existing patterns/introducing any breaking changes as much as possible

stefanlogue:
Is opening followed links in splits within the scope of this, or is that functionality you'd rather not have?

stefanlogue:
I've pushed changes to match the implementation above, and decided that opening links in a split is a possible future feature but not needed at this time

epwalsh (approved):
LGTM! Thanks @stefanlogue :)

lua/obsidian/command.lua
```diff
@@ -70,12 +72,13 @@ end
 command.new = function(client, data)
   ---@type obsidian.Note
   local note
+  local open_in = util.get_open_strategy(client.opts.open_notes_in)
   if data.args:len() > 0 then
-    note = client:new_note(data.args)
```
lucidph3nx, line 74: I suspect the change from `data.args` to `data.args[1]` has just broken the functionality of the `:ObsidianNew` command. If I call `:ObsidianNew Example`, then:
data.args = "Example"                                                                                                      
data.args[1] = nil
which breaks `note_id_func` since it never recieves the title.

stefanlogue: Ah that's a remnant of my first pass when I tried to do this via additional arguments, must have missed this change, will open a PR for a fix

Issue #167, opened 2023-08-05
ObsidianNew command doesnt read args


lucidph3nx:
### 🐛 Describe the bug

When running the command `:ObsidianNew Example` the note_id_func does not receive the 'Example' title, and only ever receives `nil`.

minimal note_id_func to reproduce:
```lua
note_id_func = function(title)
  print(title)
end,
```

I suspect this bug was caused by the change in #166 as the command.new function now looks for `data.args[1]` instead of `data.args`

### Versions

Obsidian v1.12.0

epwalsh:
Closed via #169.

PR #169, opened 2023-08-05
fix: new note args

stefanlogue:


stefanlogue:
Addresses #167

epwalsh (approved):
Thanks for the quick fix!

PR #181, opened 2023-09-03
Allow spaces in header portion of links

mscott99:
I had trouble following links, turns out it was because of a lack of the space character in a regex.


epwalsh (approved):
LGTM! Thanks!

PR #183, opened 2023-09-06
ignore '^' in link following, use %s instead to match spaces

MohamedOsman1998:
ignore '^' in link following, so that links to blocks also work as expected.
 use %s instead of ' ' to match spaces, which is the correct way mentioned in the docs, and fixes an existing problem in the previous match expression.


epwalsh (approved):
LGTM! Thanks @MohamedOsman1998

PR #186, opened 2023-09-06
Add option to limit the number of completion suggestions

### LINKED ISSUE #185: `max_item_count` for files

Currently, I was not able to find a way to limit the max number of files suggested by the cmp. I have tried adding `max_item_count` in cmp settings and it wasn't working.

Can we please add a way to limit the number of files suggested by the autocomplete, thanks! It is a lovely plugin!

### Alternatives

_No response_

### Additional context

_No response_

epwalsh:
Closes #185


epwalsh:
@mel10c let me know if this works for you. I'll hold off on merging until I hear from you.

mel10c:
Hi @epwalsh, thanks for the update. However, after trying, the `max_suggestion config` config does not seem to work for me. When the value is set to `nil`. The suggestions works perfectly, however, when its set to any integer, there ends up to be no suggesting showing after all. Only with the option to create a new note, showing in the screenshots. 

Integer 
Screenshot 2023-09-10 at 21 04 09

Versus `nil`
Screenshot 2023-09-10 at 21 04 30

epwalsh:
@mel10c I found the issue and should be fixed now. Can you try again?

mel10c:
Perfect! Its working beautifully right now!!!
Screenshot 2023-09-12 at 09 51 50

epwalsh:
@mel10c the problem is that `nvim-cmp` handles the sorting / ranking of results. As an `nvim-cmp` plugin, Obsidian just hands unsorted matches over to `nvim-cmp`. So when we truncate the matches we're not necessarily sending the best matches to `nvim-cmp`.

I think the only way to get the results you want would be to re-implement the sorting / filtering logic of `nvim-cmp`, and I really don't want to do that.

I feel like `nvim-cmp` should have a setting for this but as far as I can tell it doesn't :/

mel10c:
Oh I see, thank you so much for your help anyway!!! It has already being really helpful for me!

epwalsh:
Closing for now. This seems like a reasonable feature to request of `nvim-cmp`.

Issue #176, opened 2023-08-19
Possible encoding issue with ObsidianFollowLink


ErenKyojin:
### 🐛 Describe the bug

I'm not sure if this is a bug, an inetnded feature, or something which might be editable in settings but i have a really big vault with a lot of links i wrote "manually" with different cases compared to the actual file name, the issue is that while obsidian opens the correct file without any issue, in neovim i get a new file generated in the root folder every time.

### Versions

```
~ nvim --version
NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3
```

```
~ nvim --headless -c 'lua print("Obsidian v" .. require("obsidian").VERSION)' -c q
Obsidian v1.12.0
```

epwalsh:
It's not intended to be case sensitive, and it works as expected for me. Can you give me a particular example (note name and link) that doesn't work for you?

ErenKyojin:
Actually the issue might be with aliases, not with case sensitivity, and maybe there's also the fact that they are in sub-folders
Running ObsidianFollowLink on the link `[[Tensione|differenza di potenziale]]`, which should bring me to the file `vaultroot/elettrotecnica/tensione.md`, creates a new file `vaultroot/tensione.md`.
The same happens for, as another example, the link `[[campi di forze conservative|campo conservativo]]` creates `vaultroot/campi di forze conservative` instead of opening `vaultroot/fisica/campi di forze conservative`.

Sorry about the confusion and the late reply.

epwalsh:
Hey @ErenKyojin, thanks for the additional info. Unfortunately I still wasn't able to reproduce the issue. If you get a chance to try to debug this yourself, the relevant logic is all in here:
https://github.com/epwalsh/obsidian.nvim/blob/60f8da45f87576504e280efd0b0aa6a7d1ac6452/lua/obsidian/command.lua#L544-L597

lpanebr:
Indeed the second example debunks the case sensitivity add a culprit...

I remember having this issue also but I'm very behind on my taking notes habit..

Also I'm avoiding the computer due to lower back pain ATM.

From the behavior and code above it's clear that the `util.find_note(client.dir, note_file_name)` is not returning anything as `#notes < 1` is true.


I've reviewed the code from `util.find_note` to plenary s `scan_dir` and `is_file` and all seems perfect. I reached my limit at `Path:_stat()` in
[...]

PR #206, opened 2023-10-02
fix runtime errors on validation errors in note.from_lines

oflisback:
This PR fixes a number of runtime problems related to validation errors when reading note data in `note.from_lines`.

Prior to this fix navigating to a note with validation errors would silently fail and instead a new note would be created in the vault root. I this could also be the root cause of https://github.com/epwalsh/obsidian.nvim/issues/176.


epwalsh (approved):
Thanks @oflisback!

PR #239, opened 2023-11-20
Fix windows systems

### LINKED ISSUE #238: enum rename causing issue for Windows/WSL systems
Hi,

It seems that a recent commit #235 is causing the plugin to break on Windows systems.
Specifically, the enum `util.OSType` was renamed but the variable name wasn't changed in the subsequent OS check.
This is causing a nil reference error on line 26/31 of util.lua.

https://github.com/epwalsh/obsidian.nvim/blame/349960413adfe57a673c372d5829931730e3c9ef/lua/obsidian/util.lua#L26-L31

rolzy:
Fix #238

### DIFF
--- a/lua/obsidian/util.lua
+++ b/lua/obsidian/util.lua
@@ -23,12 +23,12 @@ util.get_os = function()
 
   local this_os
   if vim.fn.has "win32" == 1 then
-    this_os = util.OS.Windows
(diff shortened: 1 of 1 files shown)


epwalsh (approved):
Good catch, thanks @rolzy!

PR #242, opened 2023-11-22
feat: support for inline text highlighting

cdmill:
Added extmarks for inline text highlighting. I made the default highlight color the same as the highlight for the dark mode theme in the obsidian app. Feel free to squash when merging if you want to ignore my numerous index related commits.


lua/obsidian/ui.lua
```diff
@@ -391,13 +391,65 @@ local function get_line_ref_extmarks(marks, line, lnum, ui_opts)
   return marks
 end
 
+---@param marks ExtMark[]
+---@param lnum integer
+---@param ui_opts obsidian.config.UIOpts
+---@return ExtMark[]
+local function get_line_highlight_extmarks(marks, line, lnum, ui_opts)
+  if string.match(line, "==%w+==") then
```
epwalsh, line 399: It looks like this won't capture multiple space-separate words, or multiple highlights on a single line. I think you could make use of our `search.gfind()` function to make this more robust. For example:

```lua
local search = require "obsidian.search"

local s = "hi there ==highlight this text please!=="

for m_start, m_end in search.gfind(s, "==[^=]+==") do
  print(string.sub(s, m_start, m_end))
end
```

Also take a look at how [`search.find_refs()`](https://github.com/epwalsh/obsidian.nvim/blob/c1e12764bf8cd80f37a3e90cda2e44f91332709a/lua/obsidian/search.lua#L63) is implemented. The important thing in that function is that we skip matching a pattern if it's inside an inline code block.

I'm using the pattern `"==[^=]+=="`, which might be too liberal. We could always restrict that to words, numbers, whitespace, and a subset of punctuation. But it might be fine as is.

epwalsh (changes requested):
Thanks @cdmill, this is a great start! I have one suggestion on how to make the pattern matching more robust, see comment below.

lua/obsidian/ui.lua
cdmill: Actually I think the `==[^=]+==` pattern works perfectly, thanks (still getting used to Lua's regex madness..)! 

So what do you think about implementing a `search.find_highlights()` function then? One that is similar to `find_refs()` that ignores the matches inside code blocks? Or do you think it'd be better to just do that checking in `get_line_highlight_extmarks()`?

epwalsh: I think having its own function like `search.find_highlights()` would be good. Maybe you can generalize the logic in `find_refs()` into a base function that `find_refs()` and `find_highlights()` both call?

cdmill: Sounds good I'll work on that

cdmill: So I currently am still using the `==[^=]==` pattern, the only problem is it allows whitespace like `== ==`, `==text ==`, `== text==`, or `==        ==`. Unfortunately, Lua's pattern matching makes this harder than it could be. I think I would end up needing to write another function to check for sub patterns (words, etc.). Thoughts?

epwalsh: Hmm alright in that case I would suggest filtering the results of `find_matches` within `find_highlights`. Something like this:

```lua
M.find_highlight = function(s)
  local util = require "obsidian.util"
  local matches = {}
  for match in util.iter(M.find_matches(s, { "Highlight" })) do
    local match_start, match_end, m_type = unpack(match)
    local text = string.sub(match_start + 2, match_end - 2)
    if util.strip_whitespace(text) == text then
      matches[#matches+1] = match
    end
  end
end
```

This should filter out those cases.

cdmill: Perfect, that worked great. Thank you!

epwalsh (approved):
This is great! Thank you @cdmill :)

Issue #190, opened 2023-09-09
Multi vault support


mertnuhoglu:
### 🚀 The feature, motivation and pitch

I have multiple Obsidian vaults (directories). Is it possible to add support for multiple vaults?

epwalsh:
Hey @mertnuhoglu we have a PR right now that would support this: #155

oflisback:
Very nice to see that more people are interested in this. Please critique this alternative approach:

Each buffer knows what vault it belongs to automatically by checking the path hierarchy for `.obsidian` folders, these lookups can also be cached. Basically something like what I did here but returning the vault path: [is_obsidian_vault](https://github.com/oflisback/obsidian-bridge.nvim/blob/e21b42ade811112cda18c6b4d268ff4847283950/lua/obsidian-bridge/init.lua#L14-L24).

Two benefits:

* No need to explicitly set a vault dir or workspaces in the configuration.
* Each buffer knows its context so commands can be applied in the right context automatically.

epwalsh:
@oflisback, I like the simplicity of that. But I also like how #155 gives users more control. That said, I think this functionality could be built on top of #155 easily. Once that merges we can explore this more.

epwalsh:
I'll follow up on the templates question in the PR.

epwalsh:
Closing as this is now fully supported through the `config.workspaces` option. As of #277 you can also override certain settings for each workspace.

PR #257, opened 2023-12-03
Refactor workspaces

adam-devel:
this is a draft PR to decide how to go about this.
- clients must not attach to any specific root vault
  - clients must have a set of vaults (`client.workspaces`) and a currently used vault (`client.current_workspace`).
  - `client.dir` must be removed, use `.current_workspace`
  - `workspace.path` should be of type `plenary.path` if it's going to replace `client.dir`
  - `client:vault_root` should be removed
  - `obsidian.new_from_dir` should be removed
  - get rid of `templates_dir` as well ? this needs discussion
- auto command registration and clearing should happen through the workflow module, maybe `workflow:register(callback)` and `workflow:clear()`.
- we should simplify the `detect_cwd` logic: `Workspace.new_from_cwd`, `Workspace.get_workspace_from_cwd`, and `Workspace.get_default_workspace` only call each other and are not used by any other modules, the separation of the logic into these different functions unnecessary.
- we should figure out a way to deal with templates:
  from https://github.com/epwalsh/obsidian.nvim/issues/190#issuecomment-1837511113:
  > - should templates be read at some chosen "main vault" root directory (the way it's done right now)
  > - or should it be read relative to the current workspace ?
  > - or should it be a per-workspace option: `{name="foo", path="bar", template_configs..}` ?

### DIFF
--- a/lua/obsidian/backlinks.lua
+++ b/lua/obsidian/backlinks.lua
@@ -123,7 +123,7 @@ Backlinks._gather = function(self)
   local tx, rx = channel.oneshot()
 
   -- Collect matches.
-  search.search_async(self.client.dir, "[[" .. tostring(self.note.id), opts, function(match)
+  search.search_async(self.client.current_workspace.path, "[[" .. tostring(self.note.id), opts, function(match)
     if is_valid_backlink(match) then
       local path = match.path.text
 
@@ -152,7 +152,7 @@ Backlinks._gather = function(self)
   -- Load notes for each match and combine into array of BacklinksMatches.
   local executor = AsyncExecutor.new()
   executor:map(function(path, idx)
-    local ok, res = pcall(Note.from_file_async, path, self.client.dir)
+    local ok, res = pcall(Note.from_file_async, path, self.client.current_workspace.path)
     if ok then
       out[idx] = { note = res, matches = backlink_matches[path] }
     else
--- a/lua/obsidian/client.lua
+++ b/lua/obsidian/client.lua
@@ -45,7 +45,7 @@ Client.new = function(opts)
     _, default_workspace = next(self.workspaces)
   end
 
-  self:switch_workspace(default_workspace, opts)
+  self:switch_workspace(default_workspace)
 
   -- temporary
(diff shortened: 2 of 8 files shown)


adam-devel:
it's really hard to stay within the scope of this pull request, this refactor touches almost all the other modules.
i am tempted to refactor the whole thing for my personal convenience, for example i don't use templates and i don't want to deal with them so i will just remove them.
i don't want to maintain my own fork however so i will try to contribute back as much as i can so that i don't need my own fork.

adam-devel:
>>`client:vault_root` should be removed
>
>Why should we remove that?

depending on where you want workspace management to be, the client is either going manage a single vault or is going to manage multiple vaults.

for a single vault/workspace, `client:vault_root` becomes redundant with `client.current_workspace.path`.

for multiple vaults/workspaces, it's not clear which workspace's root `client:vault_root` resolve:
  - if it resolves the current workspace's root, `current_workspace` already exists.
  - if it resolves to some other workspace's root, every workspace already has a `path` field

> Alternatively that could replace client.dir.

assuming `client.dir` is the directory of the current workspace: replacing it with a function seems unnecessary as workspaces already has a `path`

to conclude: the function may seem useful somewhere outside the client module but let's remember that we require explicit paths to vaults/workspaces during `setup`, this existing expectation already makes implicit vault discovery unnecessary imo.

epwalsh:
> for a single vault/workspace, client:vault_root becomes redundant with client.current_workspace.path.

I see. From my perspective `client:vault_root()` / `client.dir` are a convenience for accessing `client.current_workspace.path`. But I don't like how we have both `client:vault_root()` and `client.dir`. One of these should definitely be removed.

adam-devel:
That's a valid perspective, sometimes it's useful to have shortcuts like that. I thought I had a deeper misunderstanding.

epwalsh:
I'm hoping to find some time over the holiday break to continue to clean this stuff up 🤞

epwalsh:
I think the workspace API is in a much better spot now, and the library has evolved substantially since this PR anyway, so I'm going to close.

Issue #256, opened 2023-12-03
feature request: add support for mini.pick in obsidiansearch command


bassamsdata:
### 🚀 The feature, motivation and pitch

the `ObsidianSearch` command currently relies on either `telescope`, `fzf-lua`, or `fzf.vim` for fuzzy searching. i'd like to propose adding support for an alternative fuzzy search tool, `mini.pick`, which is part of the `mini.nvim` module ecosystem and is currently my preferred choice for fuzzy searching.

### Additional context

i attempted to implement this enhancement through a pull request, and while i successfully implemented it, the search functionality is limited to the current working directory where neovim was started. i've provided my initial implementation below, and also linked a repository where i explored multiple approaches without success.

i invite you to review my implementation and explore the provided fork for further details. 

it would be greatly appreciated if you could consider incorporating support for mini.pick into the `obsidiansearch` command.

[My fork commits](https://github.com/epwalsh/obsidian.nvim/compare/main...bassamsdata:obsidian.nvim:main)

<details>
<summary>Click to expand code in `command.lua` file</summary>

```lua
["mini.pick"] = function()
    -- Check if mini.pick is available
    local has_mini_pick, mini_pick = pcall(require, "mini.pick")
    if not has_mini_pick then
      return false
    end

    -- Set up the arguments for mini.pick
    local grep_arguments = vim.tbl_flatten {
      base_cmd,
      {
        "--color=always",
        "--",
        util.quote(data.args),
        tostring(client.dir),
      },
    }

    -- Use mini.pick's grep_live or grep picker depending on whether there are arguments
    if data.args:len() > 0 then
      mini_pick.builtin.grep({ tool = 'rg', args = table.concat(grep_arguments, " ") })
    else
      mini_pick.builtin.grep_live({ tool = 'rg', args = table.concat(grep_arguments, " ") })
    end

    return true
end
</details>
```

Thank you so much

epwalsh:
Hey @bassamsdata I'd be happy to add this. It looks like you got a good start. Can you open a PR with your changes and I'll give it a try. Maybe I can figure out how to fix the working directory issue.

bassamsdata:
PR's been created. your help is really appreciated! 
thank you.

epwalsh:
Closed via #270

PR #258, opened 2023-12-05
Adding support for Mini.Pick plugin in the command `ObsidianSearch`

bassamsdata:
related to #256. it's working but only if started neovim in the same directory of the obsidian vault though.
I need help in correcting this.
thank you

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Templates directory location follows the workspace now.
+- Fixed bug with completion when `min_chars = 1` and you start typing an empty check box.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


bassamsdata:
I'm not sure why, but changing the function style to double brackets `()` fixed it. I attempted to match the style of previous plugin functions, but that caused the problem.

The pull request is now functional.

epwalsh (commented):
Thanks @bassamsdata and sorry for the slow response. Overall looks pretty good. We also need to add this into the `ObsidianQuickSwitch` and `ObsidianTemplate` commands (both in `lua/obsidian/command.lua`). 

> it's working but only if started neovim in the same directory of the obsidian vault though.

Is this a bug with mini.pick? Might be worth opening an issue there. I think it's okay for now if we can't fix it though. We could add a disclaimer.

bassamsdata:
appreciate the review! i've included support for `obsidianquickswitch` and `obsidiantemplate`.

> Is this a bug with mini.pick? Might be worth opening an issue there. I think it's okay for now if we can't fix it though. We could add a disclaimer.

it resolved now, functioning smoothly like telescope wherever nvim is initiated. no need for a disclaimer. 

also, added `mini.pick`to the readme file.

epwalsh:
@bassamsdata I accidentally screwed up this PR while trying to push a minor README update. Luckily I still have the commit history, just on a different branch now: https://github.com/epwalsh/obsidian.nvim/pull/270

PR #270, opened 2023-12-19
Add support for mini.pick as another alternative to telescope.nvim

epwalsh:
Continuation of #258 thanks to @bassamsdata


bassamsdata:
thank you

PR #296, opened 2024-01-03
feat(QuickSwitch): create new note on C-x

sotte:
So far, this is just an outline of how to do things.

Things to consider:

- How do we want to allow to remap the keys?
- prompt_title: yes or no? Should we adjust the other titles as well?
- If there are no matches, do we automatically generate a new note on enter?
- tests
- docs


sotte:
image

image

image

lua/obsidian/commands/quick_switch.lua
```diff
@@ -12,11 +12,28 @@ return function(client, _)
...
       end
+
       -- Search with telescope.nvim
+      local custom_actions = {
+        -- https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/actions/init.lua
+        obsidian_new = function(prompt_bufnr)
+          local query = require("telescope.actions.state").get_current_line()
+          require("telescope.actions").close(prompt_bufnr)
+          local new = require "obsidian.commands.new"
+          new(client, { args = query })
```
epwalsh, line 23: This gave me the idea to make a better API for calling our commands directly. Once https://github.com/epwalsh/obsidian.nvim/pull/298/files merges we can change this to:

```suggestion
          client:command("ObsidianNew", { args = query })
```

sotte: Nice little improvement. I'll rebase and integrate it.

sotte:
> > How do we want to allow to remap the keys?
> 
> I'm thinking we could have a configuration subtable to customize each finder. For example:
> 
> ```lua
> obsidian.setup({ finders = { telescope = { ... } } })
> ```

Thinking out loud / talking this through

We have a few options on how to configure this.

1.
We can make it super configurable so that the user can configure the action on a per finder basis.
Do we really want this? Are there even users who have multiple pickers? I don't think so.

```lua
obsidian.setup {
  finders = {
    telescope = {
      new = "C-x",
      insert_link = "C-l",
      -- potentially more; can't think of anything right now :)
    },
    fzf = {
      new = "C-x",
    },
    -- rest is not supported
  },
}
```

2.
We can make it independent of the finder and just have a global configuration.
Specific pickers only support a subset of actions.

```lua
obsidian.setup {
  finder_mappings = {
    new = "<C-x>",  -- maybe this is supported by all pickers
    insert_link = "<C-x>",  -- this only by telescope
  },
}
```

3.
We don't have any configuration and just use the default mappings.
When we have more actions than right now (create a new note) we can reconsider this approach.

Either approach is fine, but I think I prefer 3 and 2 for now. 3 we already have. Extending it to 2 is not too complex.



> 
> > If there are no matches, do we automatically generate a new note on enter?
> 
> This would be nice if it's easy enough to implement.

For telescope this is not difficult. You overwrite the default action for `CR` to do something custom when there is no match, otherwise you do the default. I can give it a try.

sotte:
Also: do we want to offer the "new" action only for ObsidianQuickSwitch or also ObsidianSearch?
If we use it for ObsidianSearch as well, the same keymaps should be used IMO.

epwalsh:
> We have a few options on how to configure this.

Let's stick with (3) for now to limit the scope of the PR. We can handle customization later if there's an obvious need for it.

> Also: do we want to offer the "new" action only for ObsidianQuickSwitch or also ObsidianSearch?

I think so.

sotte:
@epwalsh ah, just saw your message when I pushed the implementation for 2. Just implemented to see how it feels. It's not too crazy, but we only really have this one thing right now.

I can go back to option 3 if you prefer. I'm fine either way.



>>    Also: do we want to offer the "new" action only for ObsidianQuickSwitch or also ObsidianSearch?
> I think so.

Will implement it now :+1:

sotte:
Here the two supported commands with a configured mapping.

image

image

Let me know if you want the mapping to be configurable or not @epwalsh .

Then I can move on and write the docs, clean up the code, etc...

lua/obsidian/commands/quick_switch.lua
sotte: done

epwalsh:
@sotte nice job! Since you already have (2) implemented let's stick with that. I'll give this PR a proper review within the next day 👍

sotte:
@epwalsh perfect! And no rush.

I'm still new to your code base (pretty readable BTW), so feel free to point out any changes you want.

epwalsh (approved):
LGTM! I just pushed a couple minor fixes. Thank you @sotte!

sotte:
Thanks for merging this! @epwalsh

PR #300, opened 2024-01-04
Integrate mini.doc to generate Lua API docs

### LINKED ISSUE #245: Publish API docs for the public Lua API
We could use the `docgen` module from tree-sitter-lua: https://github.com/tjdevries/tree-sitter-lua/tree/master#how-to-generate-documentation

epwalsh:
Closes #245

Still TODO:
- [x] Include other modules in addition to `obsidian.client`, such as `obsidian.note`.
- [x] Run in GitHub actions.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Patch plugin docs tag
         run: |
           # Replace the title tag "*obsidian.txt*" with "*obsidian.nvim*"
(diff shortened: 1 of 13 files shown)


sotte:
Given that you already have a `Makefile`, maybe also add a make target for the docs.
I personally find this quite handy.

Here is an example of the make target that also checks out the `mini.doc` dependency: https://github.com/sotte/presenting.nvim/blob/main/Makefile#L14

PR #303, opened 2024-01-05
Fix: Fixed a bug with :ObsidianFollowLink where new notes were always created at the vault's root

Marc-AnthonyG:
When creating note through `: ObsidianLinkNew` the note is created in the current buffer directory while when using `ObsidianFollowLink` on a "undefined" note and responding `Yes` to the prompt, the note was created in the vault's root. 

I opted to align the behavior with the `:ObsidianLinkNew` functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
-- Fixed a bug with :ObsidianFollowLink where new notes were always created at the vault's root.
+### Changed
+
+- Creating new notes via `:ObsidianFollowLink` now matches the behavior of `:ObsidianLinkNew`, where the new note will be placed in the same directory as the current buffer note. This doesn't affect you if you use a flat directory structure for all of your notes.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Fixed a bug with :ObsidianFollowLink where new notes were always created at the vault's root.
+
 ### Added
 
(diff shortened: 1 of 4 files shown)


Marc-AnthonyG:
It is my first time contributing to open-source so if you have any comment on the presentation or the code feel free to give it to me.

lua/obsidian/commands/follow_link.lua
```diff
@@ -45,7 +45,7 @@ return function(client, data)
         if confirmation == "y" or confirmation == "yes" then
           -- Create a new note.
           local aliases = name == location and {} or { name }
-          note = client:new_note(location, nil, nil, aliases)
+          note = client:new_note(location, nil, vim.fn.expand "%:p:h", aliases)
```
epwalsh, line 48: I recently added this field on the client. The result is the same but I think this is a little cleaner:

```suggestion
          note = client:new_note(location, nil, client.buf_dir, aliases)
```

Could you also replace the `vim.fn.expand` call in `link_new.lua` so we're consistent?

epwalsh (commented):
Hey @Marc-AnthonyG thanks for the PR! This all looks good to me, I just left one suggestion.

lua/obsidian/commands/follow_link.lua
Marc-AnthonyG: Excellent, idea. It is done!

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Fixed a bug with :ObsidianFollowLink where new notes were always created at the vault's root.
```
epwalsh, line 11:
```suggestion
### Changed

- Creating new notes via `:ObsidianFollowLink` now matches the behavior of `:ObsidianLinkNew`, where the new note will be placed in the same directory as the current buffer note. This doesn't affect you if you use a flat directory structure for all of your notes.
```

epwalsh (approved):
LGTM!

PR #305, opened 2024-01-05
Add option to configure path to Obsidian.app on MacOS

### LINKED ISSUE #304: Configure ObsidianOpen to look into a specific folder

Currently using NIX to manage my system app and home-manager creates an app inside /Applications/ in mac os.
Configuring the ObsidianOpen to point to a path specified in the config would be optimal.

Example below:
스크린샷 2024-01-05 19 34 31

### Alternatives

_No response_

### Additional context

_No response_

epwalsh:
Closes #304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `:ObsidianFollowLink` now takes an optional "open strategy" argument. For example `:ObsidianFollowLink vsplit` or `:ObsidianFollowLink vsp` opens the note in a vertical split.
 - Added client method `Client:command(...)` for running commands directly. For example: `:lua require("obsidian").get_client():command("ObsidianNew", { args = "Foo" })`.
 - Added vim docs for the Lua API. See `:help obsidian-api` or `:help obsidian.Client`.
+- Added configuration option `obsidian_app: string|Path|?` for MacOS users who have their `Obsidian.app` in a non-standard location.
 
 ## [v2.5.3](https://github.com/epwalsh/obsidian.nvim/releases/tag/v2.5.3) - 2024-01-02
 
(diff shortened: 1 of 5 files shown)


tommasobruno:
When I run the command from the neovim terminal it works fine, but when I run the ObsidianOpen it gives error:
스크린샷 2024-01-05 21 34 06
스크린샷 2024-01-05 21 34 26

tommasobruno:
Looks like the problem is within the URI set up.
스크린샷 2024-01-05 21 44 01

epwalsh:
@tommasobruno that might be a different error. Try quoting the URI.

tommasobruno:
> @tommasobruno that might be a different error. Try quoting the URI.

Other error:
스크린샷 2024-01-05 21 58 05

epwalsh:
Ok so `../bin/obsidian` seems like the wrong path.

tommasobruno:
> Ok so `../bin/obsidian` seems like the wrong path.

It's weird, because if the don't include the URI at the end, the app opens normally

epwalsh:
Very weird. 🤔

eblume:
Apologies for the double-post but I want to be sure this doesn't get lost: https://github.com/epwalsh/obsidian.nvim/issues/304#issuecomment-2270031837

In summary: I believe this may have been closed in error. The .app bundle is located within the nix store, but whether you use that .app bundle or the package-provided `obsidian` wrapper, they both work with the command detailed above.

ajmichels:
I agree that this was maybe closed prematurely. Is there another effort to resolve this in a different way? It isn't really clear why exactly this PR was closed. Seems like one person was running into an issue that might not be representative of others and the PR was closed. Please reconsider this change. Thank you.

Because of policy restrictions on my work machine I am forced to install Obsidian in `~/Applications/Obsidian.app` so I am also not able to use any of the commands that attempt to open in the Obsidian app. It seems like having the `obsidian_app` option would still be really useful.

The following command works just fine for me when I run it manually so having this option seems like it would resolve at least some issues for some people.

```
open -a ~/Applications/Obsidian.app --background 'obsidian://open?vault=Notes&file=file%20to%20open.md'
```

PR #306, opened 2024-01-05
Accept no input as a default "yes" for prompts

Mawdac:
The capital `Y` in the prompt implies that it is the default if the
user inputs nothing and presses enter (typically). I think it would
be safe to assume that should be the behavior here too, and saves
the user keystrokes.


epwalsh (approved):
Ok with me! Thanks @Mawdac

Issue #286, opened 2023-12-31
Markdown not rendering


catasaurus:
### 🐛 Describe the bug

When I create checkboxes, they simply do not render. I don't know the expected behavior for the rendering of other markdown, but I have seen screenshots of checkboxes rendered. I know I wrote them right as when I open the file in Obsidian (the desktop client), they render perfectly fine.
```id: 1704062531-GHQA
aliases:
  - test_markdown_rendering
tags: []
---

# test_markdown_rendering

- [ ] testing this still

## markdown not rendering for some reason
- [ ] item 2

### Versions

~ via 🐍 v3.9.12 
❯ nvim --version
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1700008891

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.9.4/share/nvim"

Run :checkhealth for more info

~ via 🐍 v3.9.12 
❯               
nvim --headless -c 'lua require("obsidian").info()' -c q
[obsidian.nvim (v2.5.0)] Commit SHA: 88bf9150d9639a2cae3319e76abd7ab6b30d27f0
[plenary.nvim] Commit SHA: 55d9fe89e33efd26f532ef20223e5f9430c8b0c0
[nvim-cmp] Commit SHA: 538e37ba87284942c1d76ed38dd497e54e65b891
[telescope.nvim] Commit SHA: d90956833d7c27e73c621a61f20b29fdb7122709
ripgrep 14.0.3%                                                                                                                                     

~ via 🐍 v3.9.12 
❯

**Comment by sebassdc** [comment]
Same issue here

```
❯ nvim --headless -c 'lua require("obsidian").info()' -c q
[obsidian.nvim (v2.5.0)] Commit SHA: 88bf9150d9639a2cae3319e76abd7ab6b30d27f0
[plenary.nvim] Commit SHA: 55d9fe89e33efd26f532ef20223e5f9430c8b0c0
[nvim-cmp] Commit SHA: 538e37ba87284942c1d76ed38dd497e54e65b891
[telescope.nvim] Commit SHA: 54930e1abfc94409e1bb9266e752ef8379008592
[vim-markdown] Commit SHA: 46add6c3017d3e4035dc10ffa9cb54221d8dfe1a
ripgrep 13.0.0%
```

epwalsh:
[...]

PR #307, opened 2024-01-06
Improve UX of conceallevel UI warning

larrybotha:
With the current implementation, the warning is always shown under the following conditions:

- `conceallevel` is something other than `1` or `2` globally
- the user sets `conceallevel = 1 | 2` locally for markdown files, but opens Neovim with a buffer that is not markdown

This can be cumbersome if:

- the user's config runs `obsidian.setup` in a non-Obsidian repo and `conceallevel` is not 1 or 2
- the user sets `conceallevel` for markdown files only, but opens Neovim in an Obsidian repo on a non-markdown file

This PR addresses these scenarios by:

- using an autocommand to assert `conceallevel` on `BufEnter` _only_ for markdown files
- only runs the autocommand once per call of `obsidian.setup`

ref #286 

EDIT: add friction points, add ref to issue where warning was added

### DIFF
--- a/lua/obsidian/ui.lua
+++ b/lua/obsidian/ui.lua
@@ -567,19 +567,29 @@ M.setup = function(ui_opts)
     return
   end
 
-  local conceallevel = vim.opt_local.conceallevel:get()
-  if conceallevel < 1 or conceallevel > 2 then
-    log.warn(
-      "Obsidian additional syntax features require 'conceallevel' to be set to 1 or 2, "
-        .. "but you have 'conceallevel' set to '%s'.\n"
-        .. "See https://github.com/epwalsh/obsidian.nvim/issues/286 for more details.",
-      conceallevel
-    )
-  end
-
   local group = vim.api.nvim_create_augroup("obsidian_ui", { clear = true })
 
   install_hl_groups(ui_opts)
+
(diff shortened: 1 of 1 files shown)


marcofeltmann:
Is there any way to deactivate that warning in total?
I *know* that the rendering fails and I *want* it to show the square brackets instead of the ugly icons.

I totally understand the warning itself is a pretty good idea, but I'd love something like
```lua
require'obsidian'.setup({'conceallevelwarning' = 'skip'})
```

marcofeltmann:
TBH but that might be a bug (in the plugin or in my brain)

If I have the following setup:

1. conceallevel is 0 globally
2. I have conceallevel = 2 locally for markdown files and opens Neovim with a buffer that *is* markdown

that warning appears directly after startup.

If I set conceallevel = 1 globally and conceallevel = 2 for markdown files there is no warning.

(Oh, yeah, commit e9a8fd07035b6221e55ea568faeae006a4f76fcc it is on my machine, included via nvim pack, if that might be interesting)

epwalsh (approved):
Thanks @larrybotha, LGTM

epwalsh:
@marcofeltmann you can avoid the warning by disabling our additional syntax stuff. Does that work for you?

PR #340, opened 2024-01-22
fix for powershell command for ObsidianPasteImg on wsl

HudsonMC16:
I'm no powershell expert, but on my machine, I wasn't able to get any powershell "remote" command to work with "$content" style variables. I fiddled with the powershell command until I got it working. 

One additional issue which this unearthed: it seems that only links which match a url pattern are considered "external" links, so when you "gf" on the resulting png link from this fix, you get the raw png opened up in a buffer. Probably not ideal. Can we maybe allow the use of "follow_url_func" to be expanded to any external link, and not just urls?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Fixed powershell command for :ObsidianPasteImg in wsl
+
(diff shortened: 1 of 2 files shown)


epwalsh (approved):
Thanks @HudsonMC16! I don't have the ability to test this but I'll take your word for it :)

> One additional issue which this unearthed: it seems that only links which match a url pattern are considered "external" links, so when you "gf" on the resulting png link from this fix, you get the raw png opened up in a buffer. Probably not ideal. Can we maybe allow the use of "follow_url_func" to be expanded to any external link, and not just urls?

Seems reasonable to me if there's a good OS-agnostic way to detect non-text files from within nvim (checking extension might be fine for now). I'd be happy to review a PR.

PR #358, opened 2024-01-30
Add description to toggle_checkbox keymap opts

Mawdac:
This just adds a default description for the toggle_checkbox keymap for those
using `which-key` and the like.


epwalsh (approved):
Nice, thanks @Mawdac!

Issue #361, opened 2024-01-31
Enhance toggle_checkbox function to also convert non-checkbox lines into checkboxes


anggerdeni:
### 🚀 The feature, motivation and pitch

I am new to this plugin and I think this is an awesome project. But I came across an unexpected behavior when setting it up. Following the keymap recommendation, I set `<leader>ch` to toggle checkboxes. My expectation is as follow:

1. It will toggle between checked/unchecked for an existing checkbox [it works fine]
2. It will convert a line which is not a checkbox into a checkbox [nothing happens]

Will it be a good idea to implement the second scenario?

### Alternatives

`toggle_checkbox` should automatically insert `- [ ]` to beginning of the line. Currently I'm using the following code to achieve it
```lua
    ["<leader>ch"] = {
      action = function()
        local line_num = unpack(vim.api.nvim_win_get_cursor(0)) -- 1-indexed
        local line = vim.api.nvim_get_current_line()
        print(line)
        if not string.match(line, "^%s*- %[[ x~>-]%].*") then
          line = "- [ ]" .. line
          vim.api.nvim_buf_set_lines(0, line_num - 1, line_num, true, { line })
          return
        end

        return obsidian.util.toggle_checkbox()
      end,
      opts = { buffer = true },
    },
```

epwalsh:
@anggerdeni yes this would be good to have. Would you like to make a PR?

anggerdeni:
@epwalsh sure, I've created the PR here https://github.com/epwalsh/obsidian.nvim/pull/363

PR #363, opened 2024-01-31
feat(toggle_checkbox): transform non-checkbox line into a checkbox

anggerdeni:
`toggle_checkbox` function will transform a non-checkbox line into an (unchecked) checkbox by prepending `- [ ]` to the line. It will also respect whitespaces at the beginning of the line.

Sample cases:

before
```
some line
  some line with spacing
```

after
```
- [ ] some line
  - [ ] line with spacing
```

Closes https://github.com/epwalsh/obsidian.nvim/issues/361

### DIFF
--- a/lua/obsidian/util.lua
+++ b/lua/obsidian/util.lua
@@ -478,8 +478,17 @@ end
 util.toggle_checkbox = function()
   local line_num = unpack(vim.api.nvim_win_get_cursor(0)) -- 1-indexed
   local line = vim.api.nvim_get_current_line()
-  if not string.match(line, "^%s*- %[.*") then
-    line = string.gsub(line, "^([ ]*)", "%1- [ ] ")
+
+  local checkbox_pattern = "^%s*- %[.*"
+
+  if not string.match(line, checkbox_pattern) then
+    local unordered_list_pattern = "^([ ]*)[-*+] ([^%[])"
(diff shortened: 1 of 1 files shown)


epwalsh (commented):
When testing I noticed that if the line starts with "-" (with no checkbox), this will add a 2nd "-". Can we handle that case too?

anggerdeni:
good point @epwalsh , that apply to all the other unordered list symbols, I've updated the code, the new behavior will simply change unordered list into checkbox

```
line1
 line2
  line3
- a
-  b
-   c
  - d
  -  e
  -   f
* g
*  h
*   i
  * j
  *  k
  *   l
+ m
+  n
+   o
  + p
  +  q
  +   r
```

will be transformed into
```
- [ ] line1
 - [ ] line2
  - [ ] line3
- [ ] a
- [ ]  b
- [ ]   c
  - [ ] d
  - [ ]  e
  - [ ]   f
- [ ] g
- [ ]  h
- [ ]   i
  - [ ] j
  - [ ]  k
  - [ ]   l
- [ ] m
- [ ]  n
- [ ]   o
  - [ ] p
  - [ ]  q
  - [ ]   r
```

epwalsh (approved):
LGTM! Thanks @anggerdeni

Issue #360, opened 2024-01-31
fulltext search in vault, and insert (markdown) link to selected note in buffer


tvdab:
### 🚀 The feature, motivation and pitch

I miss (or simply can't find?) the possibility to do a full text search in the obsidian vault (with Telescope), and then insert a (markdown) link in the current buffer that points to the note that I have selected (in Telescope).  

### Alternatives

- inserting a link by typing` [[` followed by two characters in the buffer starts completion for notes in the vault, but this is not a full text search, and the preview is limited compared to what Telescope offers
- `:ObsidianSearch` offers full text search in the vault, but currently there are only two actions:
  - `<CR>`:  Open the note in a new buffer
  - `<C-x>`: Create new note

### Additional context

I hacked something together so :ObsidianSearch has an aditional action to insert a link pointing to the selected note:
![image](https://github.com/epwalsh/obsidian.nvim/assets/59047040/f6d32d91-eb61-4817-971a-26c0ab524464)

The most important change I aplied to the plugin code to achieve this, is adding an 'obsidian_insert_link' action besides the already existing 'obsidian_new' action in `picker_utils.lua`

```
M.telescope_mappings = function(map, client, initial_query)
  -- Docs for telescope actions:
  -- https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/actions/init.lua
  local telescope_actions = require("telescope.actions.mt").transform_mod {
    obsidian_new = function(prompt_bufnr)
      local query = require("telescope.actions.state").get_current_line()
      if not query or string.len(query) == 0 then
        query = initial_query
      end
      require("telescope.actions").close(prompt_bufnr)
      client:command("ObsidianNew", { args = query })
    end,

    obsidian_insert_link = function(prompt_bufnr)
      local selected_path = require("telescope.actions.state").get_selected_entry().path
      local vault_relative_path = client:vault_relative_path(selected_path)
      require("telescope.actions").close(prompt_bufnr)
      vim.api.nvim_put({"[](" ..  vault_relative_path  .. ")"}, "", false, true)
    end,
  }

  local new_mapping = client.opts.finder_mappings.new
  if new_mapping ~= nil then
    map({ "i", "n" }, new_mapping, telescope_actions.obsidian_new)
  end

  local insert_link_mapping = client.opts.finder_mappings.insert_link
  map({ "i", "n" }, insert_link_mapping, telescope_actions.obsidian_insert_link)

  return true
end
```

[...]

PR #364, opened 2024-02-01
insert_link action

tvdab:
Minimal implementation for feature request #360, probably not production ready.

### DIFF
--- a/lua/obsidian/picker_utils.lua
+++ b/lua/obsidian/picker_utils.lua
@@ -16,12 +16,25 @@ M.telescope_mappings = function(map, client, initial_query)
       require("telescope.actions").close(prompt_bufnr)
       client:command("ObsidianNew", { args = query })
     end,
+
+    obsidian_insert_link = function(prompt_bufnr)
+      local selected_path = require("telescope.actions.state").get_selected_entry().path
+      local vault_relative_path = client:vault_relative_path(selected_path)
+      require("telescope.actions").close(prompt_bufnr)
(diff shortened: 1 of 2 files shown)


lua/obsidian/picker_utils.lua
```diff
@@ -16,12 +16,25 @@ M.telescope_mappings = function(map, client, initial_query)
       require("telescope.actions").close(prompt_bufnr)
       client:command("ObsidianNew", { args = query })
     end,
+
+    obsidian_insert_link = function(prompt_bufnr)
+      local selected_path = require("telescope.actions.state").get_selected_entry().path
+      local vault_relative_path = client:vault_relative_path(selected_path)
+      require("telescope.actions").close(prompt_bufnr)
+      vim.api.nvim_put({ "[](" .. vault_relative_path .. ")" }, "", false, true)
```
epwalsh, line 24: We should allow users to customize the form of these links. We already have some relevant configuration options in the `completion` settings, such as `completion.use_path_only`, `completion.prepend_note_path`, and `completion.prepend_note_id` (see https://github.com/epwalsh/obsidian.nvim/blob/7d6574dcb8c3c7051f263824f1fdfdb1798c8de2/lua/cmp_obsidian.lua#L58-L80 for how these are used). We may need another option (eg. `completion.preferred_link_style = "markdown" | "wiki"`) in this case to switch between markdown-style vs wiki-style links.

epwalsh (commented):
Hey @tvdab, thanks for this. This is a good start, my only concern is allowing users to customize the form of the link to insert. See my comment below.

PR #367, opened 2024-02-02
insert_link action

tvdab:
Minimal implementation for feature request #360, probably not production ready.
I closed my previous related pull request #364 because I forgot to create a feature branch for this, sorry.

Not sure if the following is necessary: `local note_id = require('obsidian').Note.from_file(rel_path .. ".md", client.dir).id`. I think in most cases the id of a note is just the part of the filename after the last `/` without the extension? So maybe a regex would do to extract the id from the filename?

To be honest, I'm just starting to explore the concept of note taking in Obsidian, so I'm not sure the feature in this pull request is added value. Feel free to change/close if necessary.

### DIFF
--- a/lua/obsidian/config.lua
+++ b/lua/obsidian/config.lua
@@ -217,7 +217,7 @@ config.FinderMappingOpts = {}
 config.FinderMappingOpts.default = function()
   return {
     new = "<C-x>",
-    insert_link = "<C-i>",
+    insert_link = "<C-l>",
   }
 end
 
--- a/lua/obsidian/config.lua
+++ b/lua/obsidian/config.lua
@@ -175,6 +175,7 @@ config.LinkStyle = {
 ---@field prepend_note_id boolean
 ---@field prepend_note_path boolean
 ---@field use_path_only boolean
+---@field perferred_link_style obsidian.config.LinkStyle
 config.CompletionOpts = {}
 
 ---Get defaults.
@@ -188,6 +189,7 @@ config.CompletionOpts.default = function()
     prepend_note_id = true,
     prepend_note_path = false,
     use_path_only = false,
+    preferred_link_style = config.LinkStyle.wiki,
   }
 end
 
@@ -207,6 +209,7 @@ end
 
 ---@class obsidian.config.FinderMappingOpts
 ---@field new string|?
+---@field insert_link string|?
 config.FinderMappingOpts = {}
 
 ---Get defaults.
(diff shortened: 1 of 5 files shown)


lua/obsidian/picker_utils.lua
```diff
@@ -16,12 +16,39 @@ M.telescope_mappings = function(map, client, initial_query)
...
+      local note_id = require('obsidian').Note.from_file(rel_path .. ".md", client.dir).id
+      if client.opts.completion.preferred_link_style == "markdown" then
+        link = "["..note_id.."](" .. rel_path .. ".md)"
+      else
+        if client.opts.completion.prepend_note_id then
+          link = "[[" .. note_id .. "]]"
+        else
+          link = "[[" .. rel_path .. "]]"
+        end
+      end
```
epwalsh, line 37: In 27bc40c I added a help function to format links, so you can replace this with:

```suggestion
      local note = require("obsidian").Note.from_file(path, client.dir)
      local link = client:format_link(note)
```

lua/obsidian/picker_utils.lua
```diff
@@ -16,12 +16,26 @@ M.telescope_mappings = function(map, client, initial_query)
       require("telescope.actions").close(prompt_bufnr)
       client:command("ObsidianNew", { args = query })
     end,
+
+    obsidian_insert_link = function(prompt_bufnr)
+      require("telescope.actions").close(prompt_bufnr)
+      local path = require("telescope.actions.state").get_selected_entry().path
+      local note = require("obsidian").Note.from_file(path, client.dir)
+      local link = client:format_link(note, {})
```
tvdab, line 24: Had to pass {} to the format_link function to prevent a nil exception. Maybe this parameter should be made optional in the format_link function.

lua/obsidian/picker_utils.lua
tvdab: A good idea to provide a format_link function. I changed my commit so it makes use of this function (and also the new config enum obsidian.config.LinkStyle.

lua/obsidian/config.lua
```diff
@@ -207,13 +209,15 @@ end
...
 ---@field new string|?
+---@field insert_link string|?
 config.FinderMappingOpts = {}
 
 ---Get defaults.
 ---@return obsidian.config.FinderMappingOpts
 config.FinderMappingOpts.default = function()
   return {
     new = "<C-x>",
+    insert_link = "<C-i>",
```
sotte, line 220: I would chose a different mapping. `<C-i>` is the same as `TAB` and can cause surprising behaviour. 

Maybe `C-l` for **link**?

Just my 2 cents.

tvdab: Thx for the suggestion. Replaced <C-i> with <C-l>.

epwalsh (approved):
LGTM, thanks @tvdab!

PR #385, opened 2024-02-09
🐛 Only update frontmatter if it has changed

ianhomer:
Fixes https://github.com/epwalsh/obsidian.nvim/issues/386

With frontmatter support enabled, the undo history is populated with empty changes which are the updates to the YAML frontmatter with no change. This causes vim undo to not operate as expected.

This fix only makes the frontmatter change if there is actually a change to make.


epwalsh:
@ianhomer oops, sorry I didn't see your PR first before I pushed that fix.

ianhomer:
Thx @epwalsh - thanks for the rapid fix, great plugin BTW.

PR #433, opened 2024-02-24
FreeBSD support for ObsidianOpen

xorander00:
- Added OSType enum value for FreeBSD.
- Added FreeBSD to existing Linux platform check conditional.


epwalsh (approved):
LGTM, thanks @xorander00

PR #435, opened 2024-02-25
fix: os select bug in open command

OliverChao:
Hi, there is one very small bug that `obsidian.nvim` always chooses `util.OSType.FreeBSD` whatever your os is.


epwalsh (approved):
Good catch @OliverChao, thank you!

PR #442, opened 2024-02-27
feat: let user control whether paste image is confirmed

aaronmcadam:
This makes the process a bit faster and aligns better with Obsidian behaviour.

I've raised this PR, not expecting it to be merged, but to discuss whether it'd be helpful. 

What do you think @epwalsh?

### DIFF
--- a/lua/obsidian/img_paste.lua
+++ b/lua/obsidian/img_paste.lua
@@ -84,13 +84,20 @@ local function save_clipboard_image(path)
   end
 end
 
----@param fname string|?
----@param default_dir obsidian.Path|string
----@param default_name string|?
----@param should_confirm boolean|?
+--- @param opts table The options table containing:
+---     fname string|? The filename.
+---     default_dir obsidian.Path|string The default directory.
+---     default_name string|? The default name.
+---     should_confirm boolean|? Flag to confirm before proceeding.
 ---
----@return obsidian.Path|? image_path the absolute path to the image file
-M.paste_img = function(fname, default_dir, default_name, should_confirm)
+--- @return obsidian.Path|? image_path The absolute path to the image file.
+M.paste_img = function(opts)
+  opts = opts or {}
+  local fname = opts.fname
+  local default_dir = opts.default_dir
+  local default_name = opts.default_name
+  local should_confirm = opts.should_confirm
+
   if not clipboard_is_img() then
     log.err "There is no image data in the clipboard"
     return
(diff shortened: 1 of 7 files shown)


lua/obsidian/img_paste.lua
```diff
@@ -87,9 +87,10 @@ end
 ---@param fname string|?
 ---@param default_dir obsidian.Path|string
 ---@param default_name string|?
+---@param should_confirm boolean|?
 ---
 ---@return obsidian.Path|? image_path the absolute path to the image file
-M.paste_img = function(fname, default_dir, default_name)
+M.paste_img = function(fname, default_dir, default_name, should_confirm)
```
epwalsh, line 93:
```suggestion
M.paste_img = function(opts)
```

lua/obsidian/img_paste.lua
```diff
@@ -87,9 +87,10 @@ end
 ---@param fname string|?
 ---@param default_dir obsidian.Path|string
 ---@param default_name string|?
+---@param should_confirm boolean|?
```
epwalsh, line 90: All 4 `@param` annotations would be condensed to this:
```suggestion
---@param opts { fname: string|?, default_dir: obsidian.Path|string|?, default_name: string|?, should_confirm: boolean|? }|?
```

lua/obsidian/img_paste.lua
epwalsh: To ensure the table is never `nil`, you could add `opts = opts or {}` at the start of the function.

epwalsh (commented):
Hey @aaronmcadam, this is definitely a good addition! I just have a couple minor comments:
- Could you change the config field name to `confirm_img_paste`?
- The `paste_img` function has more positional arguments than I would like now. It would be good to condense these all into a table of options. See my suggestion below.

lua/obsidian/img_paste.lua
aaronmcadam: Thanks! I've made these changes!

lua/obsidian/img_paste.lua
aaronmcadam: Thanks!

aaronmcadam:
@epwalsh Thanks for the review!

What do you think of letting users avoid this Input dialog too?

CleanShot 2024-02-28 at 11 32 08@2x

This would make pasting images even faster.

epwalsh (approved):
LGTM, thanks @aaronmcadam!

PR #447, opened 2024-03-01
Picker improvements

### LINKED ISSUE #359: Enhanced Tag Management

While authoring new notes, determining the most appropriate tags can be challenging, particularly when the potential for duplicating existing tags exists. Current autocompletion features, although valuable, offer limited assistance, as they primarily facilitate the discovery of tags starting with similar prefixes. This limitation makes it difficult to efficiently organize and categorize notes, especially when similar but not identical tags (e.g., `my_allergies` vs. `allergies`) are in use.

## Proposed Solution
Implement an enhanced tag management feature that provides a comprehensive overview of all existing tags within the current workspace. This feature should enable users to:

1. **View a List of All Tags**: Offer an easily accessible interface that displays all tags currently in use, potentially leveraging fuzzy search to help find similar tags even if they don't share the same prefix.
2. **Tag Selection and Application**: Allow users to select multiple tags from this overview and apply them directly to the current note. This could streamline the process of tagging, ensuring consistency and preventing the proliferation of nearly duplicate tags.

## Implementation Consideration
I guess using a custom telescope picker would be an ideal approach for this feature, given its powerful search and UI capabilities such as listing tags, fuzzy searching, and multi-selection support. However, I am not sure how to actually build it.

### Alternatives

_No response_

### Additional context

_No response_

epwalsh:
Various improvements to the `Picker` API along with corresponding improvements `:ObsidianTags` command.

This command now has mappings for tagging the current note and inserting a tag. For pickers that allow it (telescope and fzf-lua) you can also select multiple tags at once and apply the mappings.

Closes #359.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added config option `picker.tag_mappings`, analogous to `picker.note_mappings`.
+
+### Changed
+
+- Renamed config option `picker.mappings` to `picker.note_mappings`.
+
 ## [v3.6.1](https://github.com/epwalsh/obsidian.nvim/releases/tag/v3.6.1) - 2024-02-28
 
 ### Added
--- a/lua/obsidian/client.lua
+++ b/lua/obsidian/client.lua
@@ -755,7 +755,7 @@ end
 ---@return obsidian.Note|?
 ---@diagnostic disable-next-line: unused-local
 Client.current_note = function(self)
-  if vim.bo.filetype ~= "markdown" then
+  if vim.bo.filetype ~= "markdown" or not self:path_is_note(vim.api.nvim_buf_get_name(0)) then
     return nil
   end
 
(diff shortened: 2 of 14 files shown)


epwalsh:
Thanks for the kind words @panosz :)

I just added a demo here: https://github.com/epwalsh/obsidian.nvim/discussions/450

I hope that clarifies things. I'm using telescope but most of that functionality is the same with fzf-lua. However with mini.pick I was not able to implemented nearly as much.

armaninux:
hi @epwalsh, long time user here, I recently moved to Telescope and I'm using its theme feature (specifically, theme = "dropdown") for every Telescope command. Not every obsidian.nvim command respect this choice, though, for example ObsidianSearch and ObsidianBacklinks. Could this be a bug/is there a setting for that? Thanks!

epwalsh:
Hey @armaninux I think this should be an easy fix. Can you show me your telescope config so I can try to reproduce?

armaninux:
Sure, I just have this one setting in my init.vim, the rest is vanilla telescope:
```
" telescope
lua <<EOF
require('telescope').setup({
  pickers = {
    find_files = {
      theme = "dropdown",
    }
  }
})
EOF
```
Single mappings can also specify that:
```
map gr :Telescope oldfiles theme=dropdown<cr>
```

epwalsh:
Ah, okay this actually does work out-of-the-box for me with our `find_files` commands like `:ObsidianQuickSwitch`. But `:ObsidianSearch` uses `grep_string` or `live_grep`, so your telescope config should look like this:

```lua
        find_files = {
          theme = "dropdown",
        },
        grep_string = {
          theme = "dropdown",
        },
        live_grep = {
          theme = "dropdown",
        },
```

On the other hand, `:ObsidainBacklinks` uses the picker API directly, but I've just pushed a fix for that to respect the theme as well: c57b248

armaninux:
Indeed it works perfectly now, nice and clean! Thanks!

PR #472, opened 2024-03-06
provide consistency between daily notes frontmatter with/without temp…

branflakes2:
Previously when opening a daily note using a template which does not include frontmatter, the default note frontmatter would be used. To me, it feels more consistent to use the default daily frontmatter in this case.


epwalsh (approved):
Good catch, thanks @branflakes2!

PR #474, opened 2024-03-06
Mark configuration fields optional to make lua lsp's diagnostics happy

ekropotin:
# Problem

A user of the library receives warnings from LUA LSP's diagnostic, about some configuration parameters are missing, even tho they are not strictly required

image

# Solution
Annotate configuration fields as optional

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ Full changelog below 👇
 - Make tags completion more efficient (less CPU time!).
 - Added file/directory completion to image name prompt from `:ObsidianPasteImg`.
 - Handle prompt cancellation gracefully.
+- Fix lua diagnostic warnings about missing fields in the configuration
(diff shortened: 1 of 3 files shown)


epwalsh (changes requested):
Hey @ekropotin, thanks for the PR. Those fields are in fact required in the *normalized* config object, so the main issue is that the type for the `opts` parameter here:

https://github.com/epwalsh/obsidian.nvim/blob/02971d227848e311077244c9ab1a86f8a3ec4c6b/lua/obsidian/init.lua#L90-L93

Should really be `obsidian.config.ClientOpts|table`.

ekropotin:
Thanks @epwalsh for the pointer. Let me do another take on this change.

epwalsh (approved):
Thanks @ekropotin, LGTM

PR #482, opened 2024-03-09
Do not write empty metadata to note

### LINKED ISSUE #480: Don't write frontmatter to file if it is empty

Currently, the plugin exposes `note_frontmatter_func` which is of the type `fun(note:obsidian.Note):table`. The table is prepended to the current buffer on save.

Now, when the table is empty, this results in an unnecessary addition of empty frontmatter at the top of the file.

```md
---
---

Other text....
```

I think it would be more intuitive if we checked the returned table using `vim.tbl_isempty()` and only inject the frontmatter if it has some content.

### Alternatives

To not affect the defaults, if that is desirable, we could change the type of `note_frontmatter_func` to `fun(note:obsidian.Note):table?`.

The `nil` value would indicate that there is no frontmatter and that it should be removed from the current file.

### Additional context

_No response_

UtkarshVerma:
As the title says, this PR introduces changes to only write the frontmatter when it has metadata. Closes #480 .

### DIFF
--- a/lua/obsidian/note.lua
+++ b/lua/obsidian/note.lua
@@ -586,6 +586,10 @@ Note.frontmatter_lines = function(self, eol, frontmatter)
   local new_lines = { "---" }
 
   local frontmatter_ = frontmatter and frontmatter or self:frontmatter()
+  if vim.tbl_isempty(frontmatter_) then
+    return {}
+  end
+
   for line in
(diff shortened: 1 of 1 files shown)


epwalsh (approved):
LGTM, thanks @UtkarshVerma

PR #487, opened 2024-03-10
fix: renaming sends proper path to joinpath

### LINKED ISSUE #486: Rename does not work when providing path
### 🐛 Describe the bug

Given the following directory structure
image
Renaming like so won't work
image

Results in  error `filename nil`

I believe I know what the error is, so will submit PR shortly.

### Config

```lua
local vault_dir = vim.env.OBSIDIAN_VAULT
require("obsidian").setup({
  dir = vault_dir,

  -- Optional, completion.
  completion = {
    -- If using nvim-cmp, otherwise set to false
    nvim_cmp = true,
    -- Trigger completion at 2 chars
    min_chars = 2,
  },
  finder = "telescope.nvim",
  open_notes_in = "current",
  attachments = {
    img_folder = "_resources",
  },
  ui = {
    enable = false,
  },
  new_notes_subdir = "current_dir",
  mappings = {
    -- Overrides the 'gf' mapping to work on markdown/wiki links within your vault.
    ["gf"] = {
      action = function()
        return require("obsidian").util.gf_passthrough()
      end,
      opts = { noremap = false, expr = true, buffer = true },
    },
    -- Toggle check-boxes.
    ["<leader>oc"] = {
      action = function()
        return require("obsidian").util.toggle_checkbox()
      end,
      opts = { buffer = true },
    },
  },
  -- Optional, customize how note file names are generated given the ID, target directory, and title.
  ---@param spec { id: string, dir: obsidian.Path, title: string|? }
  ---@return string|obsidian.Path The full path to the new note.
  note_path_func = function(spec)
    -- This is equivalent to the default behavior.
    local path = spec.dir / tostring(spec.title)
    return path:with_suffix(".md")
  end,
  wiki_link_func = "prepend_note_path",
    -- In this case a note with the title 'My new note' will be given an ID that looks
    -- like '1657296016-my-new-note', and therefore the file name '1657296016-my-new-note.md'
    local suffix = ""
    if title ~= nil then
    else
      for _ = 1, 4 do
      end
    end
    return tostring(os.time()) .. "-" .. suffix
  end,
})
```

### Environment

```text
NVIM v0.10.0-dev-2539+ga69c72063
Build type: RelWithDebInfo
LuaJIT 2.1.1707061634
Run "nvim -V1 -v" for more info
```
```text
Obsidian.nvim v3.7.1 (b136c1545fb9c4888d37ba482bfc77f222bbdf52)
Status:
   buffer directory: nil
   working directory: /home/jemag/dotfiles
Workspaces:
  ✓ active workspace: Workspace(name='obsidian', path='/home/jemag/H/obsidian', root='/home/jemag/H/obsidian')
Dependencies:
   plenary.nvim: f7adfc4b3f4f91aab6caebf42b3682945fbc35be
   nvim-cmp: 04e0ca376d6abdbfc8b52180f8ea236cbfddf782
   telescope.nvim: 7472420f8734c710bd7009081cef9b97f08a3821
Integrations:
   picker: TelescopePicker()
   completion: enabled (nvim-cmp)
Tools:
   rg: ripgrep 14.1.0
Environment:
   operating system: Linux
Config:
   notes_subdir: nil%

```

jemag:
Fixes #486

Currently self.filename is `nil` because string is passed as self.

### DIFF
--- a/lua/obsidian/commands/rename.lua
+++ b/lua/obsidian/commands/rename.lua
@@ -93,8 +93,7 @@ return function(client, data)
   local new_note_path
   if #parts > 1 then
     parts[#parts] = nil
-    new_note_path =
-      client.dir.joinpath(unpack(vim.tbl_flatten { tostring(client.dir), parts, new_note_id })):with_suffix ".md"
+    new_note_path = client.dir.joinpath(client.dir, unpack(vim.tbl_flatten { parts, new_note_id })):with_suffix ".md"
   else
     new_note_path = (dirname / new_note_id):with_suffix ".md"
   end
--- a/lua/obsidian/commands/rename.lua
+++ b/lua/obsidian/commands/rename.lua
@@ -93,7 +93,7 @@ return function(client, data)
   local new_note_path
   if #parts > 1 then
     parts[#parts] = nil
-    new_note_path = client.dir.joinpath(client.dir, unpack(vim.tbl_flatten { parts, new_note_id })):with_suffix ".md"
+    new_note_path = client.dir:joinpath(unpack(vim.tbl_flatten { parts, new_note_id })):with_suffix ".md"
   else
     new_note_path = (dirname / new_note_id):with_suffix ".md"
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Fixed renaming when passing file path
+
 ## [v3.7.1](https://github.com/epwalsh/obsidian.nvim/releases/tag/v3.7.1) - 2024-03-09
 
 ### Fixed


lua/obsidian/commands/rename.lua
```diff
@@ -93,8 +93,7 @@ return function(client, data)
   local new_note_path
   if #parts > 1 then
     parts[#parts] = nil
-    new_note_path =
-      client.dir.joinpath(unpack(vim.tbl_flatten { tostring(client.dir), parts, new_note_id })):with_suffix ".md"
+    new_note_path = client.dir.joinpath(client.dir, unpack(vim.tbl_flatten { parts, new_note_id })):with_suffix ".md"
```
epwalsh, line 96: Ah, this should actually be called as a method:
```suggestion
    new_note_path = client.dir:joinpath(unpack(vim.tbl_flatten { parts, new_note_id })):with_suffix ".md"
```

epwalsh:
Thanks @jemag!

PR #496, opened 2024-03-16
fix: erroneous title in daily note creation from template

zDonik1:
The `{{title}}` template parameter would get incorrectly replaced with the alias, which is a big number, presumably being the creation timestamp. The expected behavior is to replace it with filename.

## Example

The template:
```md
# {{title}}

```

The created daily note (filename: 2024-03-16.md):

```md
# 1710608450

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Don't open picker for tags when there aren't any matches.
+- Fix incorrect initialization of the daily note alias to a timestamp.
 - Correctly replace the `{{title}}` template parameter with filename in daily notes.
 
 ## [v3.7.3](https://github.com/epwalsh/obsidian.nvim/releases/tag/v3.7.3) - 2024-03-13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Don't open picker for tags when there aren't any matches.
+- Correctly replace the `{{title}}` template parameter with filename in daily notes.
 
 ## [v3.7.3](https://github.com/epwalsh/obsidian.nvim/releases/tag/v3.7.3) - 2024-03-13
 
--- a/lua/obsidian/client.lua
+++ b/lua/obsidian/client.lua
@@ -1886,7 +1886,7 @@ Client._daily = function(self, datetime, opts)
   if self.opts.daily_notes.alias_format ~= nil then
     alias = tostring(os.date(self.opts.daily_notes.alias_format, datetime))
   else
-    alias = tostring(os.date("%B %-d, %Y", datetime))
+    alias = tostring(os.date("%B %d, %Y", datetime))
   end
 
   ---@type obsidian.Note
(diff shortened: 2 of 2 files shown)


epwalsh:
@zDonik1 I'm confused why your alias ends up as a timestamp and not a formatted date. Can you share your config? In particular I'm curious what you have set for `daily_notes.alias_format`.

zDonik1:
I don't have anything set for the `daily_notes.alias_format`.

Here is my entire config:

```lua
require("obsidian").setup({
    sort_by = "accessed",
    disable_frontmatter = true,
    workspaces = {
        {
            name = "personal",
            path = "~/iCloudDrive/iCloud~md~obsidian/SecondBrain",
        },
    },
    daily_notes = {
        folder = "periodic_notes",
        template = "daily_template.md",
    },
    templates = {
        subdir = "templates",
        time_format = "%X",
        substitutions = {
            ["time:HH:mm:ss"] = function()
                return os.date("%X")
            end,
        },
    },
    completion = {
        nvim_cmp = true,
        min_chars = 1,
    },
    new_notes_location = "current_dir",
        return title
    end,
    wiki_link_func = "use_path_only",
    mappings = {
        ["gf"] = {
            action = function()
                return require("obsidian").util.gf_passthrough()
            end,
            opts = { noremap = false, expr = true, buffer = true },
        },
        ["<leader>oh"] = {
            action = function()
                return require("obsidian").util.toggle_checkbox()
            end,
            opts = { buffer = true },
        },
    },
    callbacks = {
        ---@param client obsidian.Client
        ---@param workspace obsidian.Workspace
        post_set_workspace = function(client, workspace)
            client.log.info("Changing directory to %s", workspace.path)
            vim.cmd.cd(tostring(workspace.path))
        end,
    },
    ui = {
        hl_groups = {
            ObsidianTodo = { bold = true, fg = mocha.peach },
            ObsidianDone = { bold = true, fg = mocha.sapphire },
            ObsidianRightArrow = { bold = true, fg = mocha.peach },
            ObsidianTilde = { bold = true, fg = mocha.red },
            ObsidianBullet = { bold = true, fg = mocha.sky },
            ObsidianRefText = { underline = true, fg = mocha.mauve },
            ObsidianExtLinkIcon = { fg = mocha.mauve },
            ObsidianTag = { italic = true, fg = mocha.blue },
            ObsidianHighlightText = { bg = mocha.flamingo },
        },
    },
})
```

zDonik1:
I have done some tests, and it seems that the alias is getting set to a timestamp because lua doesn't understand what is going on when doing date printing. It seems like there is an extra `-` in the day section. Here:

https://github.com/epwalsh/obsidian.nvim/blob/a6b5e97234a642f9f7446091fe08638f067479f2/lua/obsidian/client.lua#L1895-L1900

When I removed it, the alias was properly getting initialized.

The only thing is that it still doesn't solve the issue with the wrong `{{title}}` replacement. That is why I believe the change in this PR is also valid.

epwalsh (commented):
Huh, the "-" must be a typo, but oddly it works fine on my computer. Fixed via 02f0136.

Anyway, I don't agree that in general the note's title should be its ID. But you can achieve the same result by setting `daily_notes.alias_format` to the same formatting string as `daily_notes.date_format`.

zDonik1:
It doesn't have to be the Id. Obsidian replaces `{{title}}` with the file name, while obsidian.nvim doesn't. Thats the core of the issue. We should make the documentation more obvious of this fact if the behavior will not be changed.

PR #497, opened 2024-03-16
Add :ObsidianCreate Command for Flexible Note Creation

jmartinn:
## Overview
This PR introduces a new command, `:ObsidianCreate [PATH] [TITLE]`, to the `obsidian.nvim` plugin. This feature enhances the note creation process by allowing users to specify both a path and an optional title for new notes. This addition aims to improve flexibility and user efficiency, especially for those managing large vaults.

## Motivation
Previously, the plugin provided the `:ObsidianNew [TITLE]` command, which created new notes in a default directory. However, users, including myself, found the need to create notes in specific directories without manually navigating or adjusting configurations. This feature was born out of that need.

## Changes
- Added a new Lua file `create.lua` under `commands` to handle the `:ObsidianCreate` command logic.
- Updated the command registry in `commands/init.lua` to include `:ObsidianCreate`.
- Added an entry in the `CHANGELOG.md` under "Unreleased" to document this addition.
- Updated the `doc/obsidian.txt` file to document the new command

## Testing
The new command has been thoroughly tested in various scenarios:
- Creating a note with just a path.
- Creating a note with both a path and a title.
- Edge cases where paths might not exist or are invalid.

In all cases, the command behaved as expected, creating notes appropriately within the specified paths and handling errors gracefully.

## Future Considerations
While the current implementation meets the immediate needs, future enhancements could include:
- Better error handling for invalid paths.
- Integration with file explorers for even smoother workflow.

Thank you for considering this contribution. I believe this feature will make `obsidian.nvim` more versatile and user-friendly, especially for users with extensive vaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- New command `:ObsidianCreate [PATH] [TITLE]` for creating notes at a specified path with an optional title, enhancing flexibility in note placement within the vault.
+
 ### Fixed
 
 - Don't open picker for tags when there aren't any matches.
--- a/doc/obsidian.txt
+++ b/doc/obsidian.txt
@@ -66,6 +66,8 @@ COMMANDS                                                   *obsidian-commands*
     alias. If not given, the note corresponding to the current buffer is opened.
 - `:ObsidianNew [TITLE]` to create a new note. This command has one optional
     argument: the title of the new note.
+- `:ObsidianNew [PATH] [TITLE]` to create a new note. This command has two
+   arguments; the desired path to put the note on and an optional title.
 - `:ObsidianQuickSwitch` to quickly switch to (or open) another note in your
     vault, searching by its name using ripgrep
     <https://github.com/BurntSushi/ripgrep> with your preferred picker (see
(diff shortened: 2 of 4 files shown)


epwalsh:
Hey @jmartinn, `:ObsidianNew` does take a path instead of a title, e.g. `:ObsidianNew notes/project/foo.md`. Does that meet your needs?

jmartinn:
Hey @epwalsh, I didn't know that. However when you create a new note using the command you just showed me; `:ObsidianNew notes/project/foo.md` you have to specify a file name. 

I have a custom configuration for both `note_id_func` and `note_path_func` that basically creates a filename for the note if the title is not specified following Zettelkasten concepts.

That function only triggers when you don't specify a title / file name so my main objective here was to create a function that takes a path and without specifying a title it would create a file in the specified path and giving the note a file name based on my `note_id_func` and `note_path_func`

Here is my config for those functions
```
    -- Optional, customize how note IDs are generated given an optional title.
    ---@param title string|?
    ---@return string
      -- In this case a note with the title 'My new note' will be given an ID that looks
      -- like '1657296016-my-new-note', and therefore the file name '1657296016-my-new-note.md'
      local suffix = ""
      if title ~= nil then
      else
        for _ = 1, 4 do
        end
      end
      return tostring(os.time()) .. "-" .. suffix
    end,

    -- Optional, customize how note file names are generated given the ID, target directory, and title.
    ---@param spec { id: string, dir: obsidian.Path, title: string|? }
    ---@return string|obsidian.Path The full path to the new note.
    note_path_func = function(spec)
      -- This is equivalent to the default behavior.
      local path = spec.dir / tostring(spec.id)
      return path:with_suffix ".md"
    end,
```
This way I can create new notes with just a path and the function that handles the file name will trigger giving it a name based on your own configuration of the plugin

jmartinn:
I was finally able to accomplish what I wanted, not sure if it has to be with the latest commits that you've done @epwalsh but anyways thank you for taking your time to answer my pr. Closing this

Issue #329, opened 2024-01-17
Smart "enter" to gf_passthrough or toggle_checkbox depending on the context


sotte:
### 🚀 The feature, motivation and pitch

Often there is an obvious action to perform depending on the context (textobject the cursor is on). Why not offer a smart mapping, maybe enter, to do the right thing based on the context: 
```
if cursor_on_link:
  gf_passthrough()
elif cursor_on_checkbox_line:
  toggle_checkbox()
```

(I'm using a similar mapping in orgmode and I love it.)

### Alternatives

- keep using the current dedicated mappings
- write your own lua function to do exactly what I suggested above

epwalsh:
Sounds like a good feature. I'm happy to accept a PR with this as a default mapping.

sotte:
My ~~dumb~~ pragmatic implementation of this. Just a mapping via lazyvim really. 

```lua
return {
  {    
  "epwalsh/obsidian.nvim",
   ...
  keys = {
      {
        "<cr>",
        function()
          local util = require("obsidian.util")
          if util.cursor_on_markdown_link(nil, nil, true) then
            vim.cmd("ObsidianFollowLink")
          else
            util.toggle_checkbox()
          end
        end,
        desc = "Smart action: follow/toggle",
        ft = "markdown",
      },
```
It's slightly annoying that one action is a command and one a lua function that actually does something.

I'll play around with this the next few days and might create a MR. But if someone else comes firs, I'm not offended.

sotte:
I'm also playing with adding "show notes with tag if cursor is on tag". Not robust yet, but quite nice.

sotte:
I wanted to add `ObsidianLink` to `SmartAction`, i.e. trigger `ObsidianLink` when I'm in visual mode. This is the code that IMO should do the trick:

```lua
ObsidianSmartAction = function(opts)
  -- opts comes from the user command
  -- https://neovim.io/doc/user/lua-guide.html#lua-guide-commands-create
  -- print(vim.inspect(opts))

  local mode = vim.api.nvim_get_mode().mode
  if mode == "v" then
    local client = require("obsidian").get_client()
    client:command("ObsidianLink", opts)
    return
  end
  local obs_util = require("obsidian.util")
  if obs_util.cursor_on_markdown_link(nil, nil, true) then
    vim.cmd("ObsidianFollowLink")
  else
    vim.cmd("SmartToggleTask")
  end
end
```

[...]

PR #518, opened 2024-03-27
Added smart action to perform depending on the cursor context (#329)

CaeChao:
### Added

- Added a smart action to perform depending on the context (textobject the cursor is on).

### Changed

* Added custom UI opts for toggle checkbox utils, and changed its behavior to cycle through checkboxes

### DIFF
--- a/lua/obsidian/commands/checkbox.lua
+++ b/lua/obsidian/commands/checkbox.lua
@@ -1,0 +1,14 @@
+local toggle_checkbox = require("obsidian.util").toggle_checkbox
+
+---@param client obsidian.Client
+return function(client)
+  local checkboxes = vim.tbl_keys(client.opts.ui.checkboxes)
+  for k, v in pairs(client.opts.ui.checkboxes) do
+    local order = v.order
+    if order and type(order) == "number" then
+      -- sort the checkboxes based on order
+      checkboxes[order] = k
+    end
+  end
+  toggle_checkbox(checkboxes)
+end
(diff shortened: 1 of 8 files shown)


lua/obsidian/commands/checkbox.lua
```diff
@@ -0,0 +1,14 @@
...
+---@param client obsidian.Client
+return function(client)
+  local checkboxes = vim.tbl_keys(client.opts.ui.checkboxes)
+  for k, v in pairs(client.opts.ui.checkboxes) do
+    local order = v.order
+    if order and type(order) == "number" then
+      -- sort the checkboxes based on order
+      checkboxes[order] = k
+    end
+  end
```
epwalsh, line 12: I think sorting here based on order would be more robust. Something like:

```suggestion
  table.sort(checkboxes, function(a, b)
    return client.opts.ui.checkboxes[a].order < client.opts.ui.checkboxes[b].order
  end)
```

epwalsh (changes requested):
Thanks for this @CaeChao! Just a few comments

lua/obsidian/commands/checkbox.lua
CaeChao: Thanks for the advice! I'll commit the changes

sotte:
Not sure if we want to add something like this, but accessing the tags via the "smart action" is quite handy IMO:

```lua
  -- show tags if possible
  local current_wrod = vim.fn.expand("<cWORD>")
  if current_wrod:match("^#") then
    vim.cmd("ObsidianTags")
    return
  end
```

@CaeChao thanks for bringing this feature over the finish line.

PR #530, opened 2024-04-05
Update `<CR>` mapping in README

parmort:
The mapping for `<cr>` in the README didn't have the `expr` option enabled for the mapping to work correctly.


epwalsh (approved):
Thank you!

Issue #534, opened 2024-04-07
A way to append to frontmatter from templates would be amazing.


Jihillestad:
### 🚀 The feature, motivation and pitch

Different type of notes sometimes need different lines in the frontmatter like f.ex **type**, **author**, **source** etc. A way to implement appending lines to the auto-generated frontmatter from templates would make this a lot more dynamic.

PR #539, opened 2024-04-10
fix: ensure template insertion to occur at the current line.

XiaowenHu96:
`nvim_buf_set_lines` function uses zero-based indexing while `nvim_win_get_cursor` returns (1,0)-indexed cursor position.
This caused templates to be inserted one line below the expected cursor position. 

Additionally, users were unable to customize the frontmatter in their templates due to it being placed on the second line, which was not recognized properly by obsidian.nvim. (#534)


epwalsh (approved):
Thanks @XiaowenHu96! LGTM

PR #548, opened 2024-04-22
Add optional line number argument to toggle_checkbox, defaulting to the current line number if not provided.

milesmmoran:
I was hoping to add an optional line argument to toggle_checkbox, so I could checkboxify lines in mass.

My lua/nvim experience is very limited. ChatGPT guided my hand. I can follow up with proper documentation, but wanted to run this by somebody to ensure i'm not duplicating effort and to verify that this is not a bad idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added an optional line number argument for toggle_checkbox, that defaults to the current line.
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added an optional line number argument for toggle_checkbox, that defaults to the current line.
```
epwalsh, line 12:
```suggestion
- Added an optional line number argument for `util.toggle_checkbox()`, that defaults to the current line.
```

epwalsh (approved):
LGTM, thanks @milesmmoran

PR #553, opened 2024-04-25
Fix config annotation for `note_frontmatter_func`

parmort:
Running the `setup` function as follows would throw an LSP warning because no return type was given in the annotation for `note_frontmatter_func`. I just added the return type of table to the field annotation.

```lua
require('obsidian').setup({
  note_frontmatter_func = function(note)
    return {}
  end
})
```


epwalsh (approved):
Thanks!

PR #590, opened 2024-05-17
Only collect anchors/blocks when prompted in completion

### LINKED ISSUE #589: For completion, default to links/aliases, make leading #/^ complete for anchors/blocks

Since `v3.7.0`, the addition of support for anchor and block search has resulted in the default behavior of the completion to change.

Previously typing `[[` would (after a configurable amount of characters) initiate search among the note names and their aliases, much like how Obsidian works. The recent addition of the support for anchors and blocks changes two things:

1. The completions no longer resemble the default behavior of Obsidian, due to Obsidian requiring a prefix after `[[` for initiating a completion different than note names and aliases (`[[#` for headings in the current note, `[[##` in all notes; `[[^` for blocks in this note, and `[[^^` in all notes).
2. The amount of completions to process in the vault increases, meaning that the time before the completion menu pops up has increased notably.

While having the ability to get completion for anchors and blocks is useful, to both be more in line with the working of Obsidian, and to avoid the performance penalty of always completing for everything (notes, aliases, anchors, and blocks), would it be a good idea to introduce a similar kind of prefix detection (i.e. looking for something like `[[#`) to the completion triggering?

### Alternatives

An alternative would be making it possible to disable the anchor and block completion, to go back to the previous way things worked. The README made me think that this should be possible by changing the `wiki_link_func` config opt, but (unless I've missed something) this does not seem to do the trick.

### Additional context

_No response_

epwalsh:
Closes #589.


epwalsh:
Hey @linduxed, moving the discussion here...

> how is one to reference headers in the same document with this PR

`[[#...` should work. It does for me.. Maybe there's something off in your config. Can you share it?

PR #593, opened 2024-05-19
fix: 0.11 compat

tris203:
I have tested this and it seems to work well for me

~~I had test issues, but i was getting them on main too~~

~~Would appreciate if anyone has ideas why. Searching seems to work fine in my tests~~

Fixed in comment below


tris203:
The fix was to add the .filename field to the search, as the new flatten function cannot flatten dict type tables

The tests now pass and everything seems to work from my testing

epwalsh (approved):
Thanks @tris203, LGTM

PR #598, opened 2024-05-25
fix: upgrade to vim.islist without errors

JojoMakesGames:
Was getting an error when starting up neovim sourced from this plugin:
image

Fix was to follow the deprecation style in the same file:
```lua
compat.flatten = function(t)
  if vim.fn.has "nvim-0.11" == 1 then
    return vim.iter(t):flatten():totable()
  else
    return vim.tbl_flatten(t)
  end
end
```
This removes the error. All tests pass since functionality is the same.

Chose not to add to changelog since this seemed to fit with the line:
"- Add compatibility for NVIM 0.11"


JojoMakesGames:
Let me know if I need to add to the changelog in this case. It felt redundant but I see that the workflow requires it for a successful PR

epwalsh (approved):
Thanks @JojoMakesGames!

PR #620, opened 2024-06-08
Make searching for notes by filename case insensitive

ibash:
I have a note named "Rob Snyder.md". When I tried to link to it with cmp like "[[rob " the note wasn't found.

After this change the note is found. Note that `ignore_case` will set the `--ignore-case` flag in the rg command. However that flag is ignored and instead `--glob-case-insensitive` is what affects results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Searching for notes by file name is case insensitive.
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Searching for notes by file name is case insensitive.
```
epwalsh, line 10:
```suggestion
### Fixed

- Searching for notes by file name is case insensitive.
```

epwalsh (approved):
LGTM, thanks @ibash!

PR #621, opened 2024-06-11
Add `:ObsidianNewFromTemplate` command

malko42:
The goal is to be able to create a new note while using an existing template


lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
+local util = require "obsidian.util"
+local log = require "obsidian.log"
+
+---@param client obsidian.Client
+return function(client, data)
+  ---@type obsidian.Note
```
epwalsh, line 6: this needs to be right above `local note` (line 12).
```suggestion
  ---@type obsidian.Note
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) with respect to the public API, which currently includes the installation steps, dependencies, configuration, keymappings, commands, and other plugin functionality. At the moment this does *not* include the Lua `Client` API, although in the future it will once that API stabilizes.
 
+### Added
+
+- Added `ObsidianNewFromTemplate` command. This command will create a new note from a template.
+
```
epwalsh, line 11: Can you move this to under `## Unreleased` below?
```suggestion
### Added

- Added `ObsidianNewFromTemplate` command. This command will create a new note from a template.

```

lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
...
+  if not client:templates_dir() then
+    log.err "Templates folder is not defined or does not exist"
+    return
+  end
+
+  local note
+  if data.args:len() > 0 then
```
epwalsh, line 13: This would make this a little more robust, in case `data.args` is `nil`.
```suggestion
  if data.args and data.args:len() > 0 then
```

epwalsh (commented):
This looks pretty good @malko42! Just a few minor comments

lua/obsidian/commands/new_from_template.lua
malko42: Done

lua/obsidian/commands/new_from_template.lua
malko42: Done

malko42:
> This looks pretty good @malko42! Just a few minor comments

Thanks, you're too kind, I just hacked this together from existing functions

lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
...
+    log.err "No picker configured"
+    return
+  end
+
+  picker:find_templates {
+    callback = function(path)
+      insert_template(path)
+    end,
+  }
+  client:write_note_to_buffer(note)
```
epwalsh, line 44: I think this is redundant since you already call this in the callback

lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
...
+
+  local function insert_template(name)
+    client:write_note_to_buffer(note, { template = name })
+  end
+
+  local picker = client:picker()
+  if not picker then
+    log.err "No picker configured"
+    return
+  end
```
epwalsh, line 37: Can you move this up to the start of the command function so that we error immediately if a picker isn't configured?

lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
...
+
+  picker:find_templates {
+    callback = function(path)
+      insert_template(path)
```
epwalsh, line 41:
```suggestion
      client:write_note_to_buffer(note, { template = name })
```

lua/obsidian/commands/new_from_template.lua
```diff
@@ -0,0 +1,45 @@
...
+    note = client:create_note { title = title, no_write = true }
+  end
+
+  -- Open the note in a new buffer.
+  client:open_note(note, { sync = true })
+
+  local function insert_template(name)
+    client:write_note_to_buffer(note, { template = name })
+  end
+
```
epwalsh, line 32: We can just move this logic directly into the callback for simplicity.
```suggestion
```

epwalsh (commented):
@malko42, sorry just noticed a few more nit-picky things on my last review. Could you also add a description of this command to the corresponding section in the README with the other commands?

Issue #411, opened 2024-02-19
Preview a Image link


csyJoy:
### 🚀 The feature, motivation and pitch

I'm not sure if this is a proper feature request, but the lack of preview image functionality in my use of this plugin has forced me to revert back to Obsidian, I know there are other ways to achieve similar functionality, such as using kitty and some of the nvim plugins for image previews, but I think it's a bit heavy handed, and the only functionality I'm looking for is when I come across an image link like `![[image.png]]`, I can use the same link jump shortcut as `gf` to open a preview window, once I have this feature, all my work can be done in the terminal ! : )

### Alternatives

using [kitty](https://github.com/kovidgoyal/kitty) + [hologram.nvim](https://github.com/edluffy/hologram.nvim) to preview all image in nvim.

### Additional context

My OS is MacOS and I use alacritty as my terminal, so  kitty + hologram.nvim solution was not on my mind, in fact I have implemented it in my own work environment using a `qlmanage` cli tool in macOS, but since this implementation is platform-specific and I don't know if this is a feature needed by other users. So I didn't submit the pr, here's a demonstration video:

https://github.com/epwalsh/obsidian.nvim/assets/58020450/57d69a2a-6b49-406a-b3ed-dfa856ddbf23

stelcodes:
Maybe this feature could implemented on Linux by introducing a command that called `xdg-open` so it could work on both `http/s` links (opening them in a browser) or images (opening them in an image preview program).

csyJoy:
> Maybe this feature could implemented on Linux by introducing a command that called `xdg-open` so it could work on both `http/s` links (opening them in a browser) or images (opening them in an image preview program). Apple has an `open` command that's similar I believe.

Thanks for your comment, the reason why I use `qlmanage ` not `open` is `open` must move your hand to your mouse to close the preview window(or use a complex shortcut), whereas `qlmanage` gives a more efficient way to close it using Space, I have a slight obsession with pure keyboard use so I choose `qlmanage`. Maybe it is also a important aspect for choosing preview tool.

epwalsh:
This is now possible on Mac OS since 47f0074. See the `follow_img_func` field in the config example in the README.

ostarJoy:
So nice! I can finally return to the main branch : )

PR #660, opened 2024-07-21
Allow customizing how markdown links are followed

ibash:
In particular this allows previewing pasted images with "gf" on mac. This implements #411


epwalsh (commented):
Hey @ibash, I'm not sure this is the right approach since many people use markdown and wiki links interchangeably. I don't think it makes sense to have a dedicated function for just markdown links.

If the goal is just to be able to preview images, I've implemented that in  47f0074.

ibash:
That's exactly what I was looking for, thanks!

PR #665, opened 2024-07-26
Issue 632 support for obsidian callouts

aquilesg:
This adds UI support for the callout groups that are defined 
in the [Obsidian](https://help.obsidian.md/Editing+and+formatting/Callouts#Supported%20types) documentation.

It also supports custom alias definitions.

The result is below:
Screenshot 2024-07-25 at 7 01 11 PM


aquilesg:
Looking at it makes me wonder if it's cleaner to not have the blue bit before the highlight group

lua/obsidian/ui.lua
```diff
@@ -515,14 +691,30 @@ local function update_extmarks(bufnr, ns_id, ui_opts)
 
     -- Check if inside a code block or at code block boundary. If not, update marks.
     if string.match(line, "^%s*```[^`]*$") then
-      inside_code_block = not inside_code_block
+      inside_code_block = true
```
epwalsh, line 694: Why this change?

lua/obsidian/util.lua
```diff
@@ -1332,4 +1332,39 @@ util.buffer_is_empty = function(bufnr)
...
+  if #stack == 0 then
+    return nil
+  end
+  return stack[#stack]
+end
+
+-- Check if the stack is empty
+function util.is_empty(stack)
+  return #stack == 0
+end
```
epwalsh, line 1377: Could you move this to `lua/obsidian/collections.lua`

lua/obsidian/ui.lua
```diff
@@ -515,14 +691,30 @@ local function update_extmarks(bufnr, ns_id, ui_opts)
...
+        for c = 1, #line do
+          local char = line:sub(c, c)
+          if char == ">" then
+            count = count + 1
+          elseif char == "[" then
+            if line:sub(c, c + 2) == "[!]" then
+              break
+            end
+          end
+        end
```
epwalsh, line 710: If this is just counting the number of ">"s then you could do that with regular expressions

lua/obsidian/search.lua
```diff
@@ -36,6 +36,7 @@ M.Patterns = {
 
   -- Miscellaneous
   Highlight = "==[^=]+==", -- ==text==
+  Callout = "%[!.*%]",
```
epwalsh, line 39: Shouldn't callouts always start with ">"?

epwalsh (changes requested):
Hey @aquilesg this is a great start and I'm really excited about this feature. I ran into some bugs when I tried it out though:

When I have this:
```md
>[!info] foo
```
it appears to work, although I think the space between ">" and "[" should be required.

image

When I do add the space:

```md
> [!info] foo
```

I don't see the info icon anymore.

image

And when I add another line the first non-space character gets swallowed regardless of whether or not I have a space after ">":

```md
> [!info] foo
> bar
```

image

lua/obsidian/ui.lua
aquilesg: Oh whoops, I didn't mean to add this. Let me revert.

lua/obsidian/search.lua
aquilesg: Ah that's right

lua/obsidian/ui.lua
aquilesg: I'm a n00b with Lua and regex so I'm struggling to find the correct way to do this. 
Do you have any tips?

aquilesg:
I think I fixed most issues.
Screenshot 2024-07-30 at 8 04 08 AM

lua/obsidian/collections.lua
```diff
@@ -91,4 +91,39 @@ end
...
+  if #stack == 0 then
+    return nil
+  end
+  return stack[#stack]
+end
+
+-- Check if the stack is empty
+function M.is_empty(stack)
+  return #stack == 0
+end
```
epwalsh, line 127: This isn't really how Lua classes are implemented, and it doesn't seem necessary to have a separate class here anyway since it's just a very thin wrapper around a raw table/list.

lua/obsidian/ui.lua
```diff
@@ -473,14 +474,226 @@ local function get_line_highlight_extmarks(marks, line, lnum, ui_opts)
...
+---@param callout_hl_group_stack {}
+---@param callout_mark_start integer|nil
+---@param callout_mark_end integer|nil
+local function generate_callout_extmarks_body(
+  marks,
+  line,
+  lnum,
+  callout_hl_group_stack,
+  callout_mark_start,
+  callout_mark_end
```
epwalsh, line 489: I don't think these args are actually used

lua/obsidian/ui.lua
epwalsh: I think you could just do `local count = util.string_count(line, "> ")`

lua/obsidian/collections.lua
aquilesg: I'm fine with removing this all together then.

PR #674, opened 2024-07-31
fix: deal with lack of opts.attachment

delphinus:
Without this fix, it causes an error when `opts.attachments` does not exist.


epwalsh (approved):
Thanks @delphinus!

PR #668, opened 2024-07-29
Add `:Obsidian [subcommand]` command

epwalsh:

PR #693, opened 2024-08-22
Also add the obsidian commands to the lazy sample in the README.

Julian:
(Otherwise as-is, with suggesting lazy = true, running e.g. :ObsidianQuickSwitch will not work until opening a markdown file.)

This may need updating after #668 is merged -- and of course you could show a shorter list of just the commands which make any sense to run when not having a note already open, but it was easier to just copy paste the list out of the README and not use my brain :)

Thanks for the plugin!


README.md
```diff
@@ -134,6 +134,30 @@ return {
   --   "BufReadPre path/to/my-vault/*.md",
   --   "BufNewFile path/to/my-vault/*.md",
   -- },
+  cmd = {
```
Julian, line 137: (https://lazy.folke.io/spec#spec-lazy-loading)

PR #699, opened 2024-08-30
fix: ensure Telescope preserves original file order

bigzhu:
Telescope was reordering files previously sorted by 'rg'


bigzhu:
feat(util): add multibyte character support for visual selection

Add isMultibyteChar function to properly handle multibyte characters 
(like CJK characters, emojis) in visual selection. This fixes potential 
issues when selecting text containing characters that occupy multiple bytes.

Issue #714, opened 2024-09-14
Frontmatter of template is not merged with existing frontmatter


meeIbrahim:
### 🐛 Describe the bug

Create a note with frontmatter and then use `:ObsidianTemplate` to insert a template with frontmatter
For reference:
A note with frontmatter 
![image](https://github.com/user-attachments/assets/6fe5b041-68e6-44ce-b7d2-be0ffd3f5494)

**What Happens**: Template is inserted at the location of cursor along with its frontmatter as a simple text
![image](https://github.com/user-attachments/assets/74711fd6-7f8b-4ad2-8edd-f2c38af97697)

**What should happen**: frontmatter from template should be merged with frontmatter of the buffer
![image](https://github.com/user-attachments/assets/b852a446-51c8-4227-80b6-3c50b78fb473)

### Config

```
return {
    -- Git URL
    "epwalsh/obsidian.nvim",
     version = "*",


    lazy = false,
    ft = "markdown",
    dependencies = {
        "nvim-lua/plenary.nvim",
    },
    config = function()
        vim.o.conceallevel = 2 -- set conceal level for obsidian plugin
        require("obsidian").setup({
            workspaces = {
                {
                    name = "Ibrahim's Vault",
                    path = os.getenv("OBSIDIAN_VAULT_DIR"),
                },
            },
            notes_subdir = "00_inbox",
            new_notes_location = "notes_subdir",

            disable_frontmatter = false,
            templates = {
                subdir = "templates",
                date_format = "%Y-%m-%d",
                time_format = "%H:%M:%S",
            },

            -- name new notes starting the ISO datetime and ending with note name
            -- put them in the inbox subdir
            -- note_id_func = function(title)
            --   local suffix = ""
            --   -- get current ISO datetime with -5 hour offset from UTC for EST
            --   local current_datetime = os.date("!%Y-%m-%d-%H%M%S", os.time() - 5*3600)
            --   if title ~= nil then
            --     suffix = title:gsub(" ", "-"):gsub("[^A-Za-z0-9-]", ""):lower()
            --   else
            --     for _ = 1, 4 do
            --       suffix = suffix .. string.char(math.random(65, 90))
            --     end
            --   end
            --   return current_datetime .. "_" .. suffix
            -- end,

            -- key mappings, below are the defaults
            mappings = {
                -- overrides the 'gf' mapping to work on markdown/wiki links within your vault
                ["gf"] = {
                    action = function()
[...]

PR #715, opened 2024-09-14
Merge frontmatter when inserting template

meeIbrahim:
Fix for issue: https://github.com/epwalsh/obsidian.nvim/issues/714

Changed insert template method to merge frontmatter of template into the existing note if note already has frontmatter


meeIbrahim:
Need to cater for disable_frontmatter. Converting to draft

meeIbrahim:
@epwalsh  We might be able to improve this merging behavior further if we change some code in the Client. Let me know if you think this is enough or should i give this a re-go

PR #728, opened 2024-09-28
Add option to generate Obsidian-compatible wiki links

BeltranLeo:
This PR adds a new option `use_name_only` for the `wiki_link_func` setting that generates links in the `[[Foo]]` format. This ensures compatibility with the Obsidian desktop app, which might not recognize links in the `[[alias]]` format.

#### Changes made:

* Added the `use_name_only` option to `wiki_link_func`, which generates links using the filename and the relative path without the extension.
* This configuration can be customized to better fit the folder structure of each user.
* The `use_name_only` option differs from the existing `use_alias_only`, as `use_alias_only` might not work correctly with the desktop version if the alias is not identical to the filename.

#### Reason for the contribution:

The default autocompletion function generated links that were not recognized by the Obsidian desktop app, making note navigation more difficult. With this function, links are compatible and easy to use in both `obsidian.nvim` and the Obsidian app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `opts.follow_img_func` option for customizing how to handle image paths.
 - Added better handling for undefined template fields, which will now be prompted for.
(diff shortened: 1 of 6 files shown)


BeltranLeo:
I've encountered an error in the tests, so I will cancel this pull request for now. I'll fix the issue and open a new PR once everything is resolved. Thank you for your understanding.

PR #729, opened 2024-09-28
Add option to generate Obsidian-compatible wiki links

BeltranLeo:
This PR adds a new option `use_name_only` for the `wiki_link_func` setting that generates links in the `[[Foo]]` format without the `.md` extension. This ensures compatibility with the Obsidian desktop app, which does not recognize links in the `[[alias]]` format.

### Changes made:

* Added the `use_name_only` option to `wiki_link_func`, which generates links using the filename and the relative path without the extension.
* The `use_name_only` option differs from the existing `use_alias_only`, as `use_alias_only` might not work correctly with the desktop version if the alias is not identical to the filename.
* Fixed the tests to ensure they correctly handle the `wiki_link_name_only` function and produce the expected results.
* The `use_name_only` option differs from the existing `use_alias_only`, as `use_alias_only` might not function correctly with the desktop version if the alias does not match the filename exactly.

### Reason for the contribution:

The default autocompletion function generated links that were not recognized by the Obsidian desktop app, making note navigation more difficult. With this function, links are compatible and easy to use in both `obsidian.nvim` and the Obsidian app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `opts.follow_img_func` option for customizing how to handle image paths.
 - Added better handling for undefined template fields, which will now be prompted for.
+- You can now set `wiki_link_func` to `use_name_only` to generate wiki links in the format `[[Foo]]`, ensuring compatibility with the Obsidian desktop app.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ This is a complete list of all of the options that can be passed to `require("ob
   --  * "prepend_note_id", e.g. '[[foo-bar|Foo Bar]]'
   --  * "prepend_note_path", e.g. '[[foo-bar.md|Foo Bar]]'
   --  * "use_path_only", e.g. '[[foo-bar.md]]'
+  --  * "use_name_only", e.g, '[[foo-bar]]'
   -- Or you can set it to a function that takes a table of options and returns a string, like this:
   wiki_link_func = function(opts)
     return require("obsidian.util").wiki_link_id_prefix(opts)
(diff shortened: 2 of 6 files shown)


tihawk:
Nice! For my use-case I also remove the path, and so I just keep the file name. I guess that'd be the real `use-name-only` xD

The reasoning is that if I move a note from outside of Obsidian Desktop (via a script for example), it would not update the file path inside of links, so it would break the links.

But yeah, I was surprised that most of the available in the plugin options for generating links are unsupported by the Obsidian app. This PR is needed IMHO.

BeltranLeo:
Could you give some examples so I can better understand your use case?

From my understanding, Obsidian uses the name of the notes (without the extension) in order to be able to locate the files. It can also use the relative path from the root of the vault to achieve the same result. For this reason, moving new files to the Obsidian directory should allow to link them from other notes without problems. Please let me know if I understand this correctly.

The other option is that you want to link to files outside the Obsidian directory, which could bring problems if you want to access the linked note from another device.

I’m also looking at adding an option to facilitate the inclusion of subsections within the note, titles or tags, but the current system uses a strange character that is not fully compatible with the standard Obsidian form (which can lead to inconsistencies if some notes are created from Obsidian and others from Neovim).

tihawk:
Sure. You're right that Obsidian uses either a filename or a relative filepath to create links. In my case I avoid including the filepath for the following reason:

My workflow:
- I automatically create new notes in a directory called `inbox` inside my vault. It doesn't matter if the new note is created with a :ObsidianNew[FromTemplate] command or by typing in a new link for a file that doesn't exist.
- I then edit the new notes, and once in a while move them to another directory, where they belong.

The crux is that I don't typically use the Obsidian app to move files around. I either do it with some bash scripts, or just from inside nvim.

So, if I had created a link in a note that is something like `[[inbox/file-name|a note]]`, and then moved the `file-name` in any other way than from the Obsidian app to say `notes/resource/file-name`, Obsidian wouldn't know to update the link inside the note that links to `file`. But if my link was simply `[[file-name|a note]]`, then nothing needs to get updated, and Obsidian is still happy to link to `file-name`, as long as it's inside the vault.

Hence, I simply added a small regex to the function I took from this PR
```lua
local name = opts.path:match("[^/]+$"):gsub("%.md", "")
```
, although I don't know if the forward-slash is platform-independent in this path. I haven't looked at how paths are parsed in the plugin.

Issue #492, opened 2024-03-13
Code blocks with conceallevel


chris-dem:
I am trying to add code blocks to my documentation but due to the conceallevel, the backticks dissapear, is there a way to tell obsidian to overlay the backtick block because I would like to keep the rest of the UI features enabled

aaronmcadam:
I'd like this feature too.

I tried to fix this myself by adding an autocommand to check if the current line contains backticks, but I couldn't get it to work correctly.

aharoJ:
Is there any progress on the backticks? I was going to open an issue until I saw this one

igorMSoares:
The workaround I'm currently using for this is to set a replacement character for the concealed triple backtick (```)
Making the code block to look like this:

![2024-06-27_13-48](https://github.com/epwalsh/obsidian.nvim/assets/108494295/8d7fda73-43c9-409e-ac42-a6b1d664992f)
![2024-06-27_13-49](https://github.com/epwalsh/obsidian.nvim/assets/108494295/96ac4eeb-a7e1-44fa-9fd1-6b28ae873943)

Still not ideal, but at least gives me an indication that this is a code block

I have configured a VimEnter autocmd for that:

```lua
vim.api.nvim_create_autocmd('VimEnter', {
  callback = function()
    vim.cmd [[
      augroup MarkdownSyntaxMatch
        autocmd!
        autocmd FileType markdown syntax match @conceal /```/ conceal cchar=⋯
      augroup END
    ]]
  end,
})

```

You can also set a different character for the triple backticks of the end of the codeblock by changing the match pattern:

```lua
syntax match @conceal /```\n/ conceal cchar=x
```

gfvioli:
I'm also suffering from this issue and I would really like a nicer solution to this problem. Maybe a way to only add concealing to your vault folder or a way to do some conceal "scaping".

igorMSoares:
@gfvioli I started noticing this issue just a couple of days after I commented here with the autocommand. Something has changed after that, that's why in the screenshot I posted the whole ` ```ts` is replaced with the replacement character. Since then I haven't had the time to figure out why the behaviour has changed or to try any workaround. Looking forward to see if someone has got any ideas about it.

nobe4:
This is most likely not an `obsidian.nvim` issue but a `nvim-treesitter` issue.

Setting `conceallevel = 2` does hide the ` ``` ` even without `obsidian.nvim` installed.

[...]

PR #778, opened 2024-11-26
feat: add an UI option to ignore conceal warnings

nobe4:
In https://github.com/epwalsh/obsidian.nvim/issues/286#issuecomment-2048334358 it was suggested that a new config could be added to specifically not warn if the conceal level is 0 and the UI is enabled.

This commit does this and guard against the conceal warning when the option is set.

cc https://github.com/epwalsh/obsidian.nvim/issues/492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Added `opts.ui.ignore_conceal_warn` option to ignore conceal-related warnings.
+
 ### Added
 
 - Added `opts.follow_img_func` option for customizing how to handle image paths.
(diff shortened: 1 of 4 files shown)


kmac:
I would love to see this go in.  I do want the UI decorations that are applied with obsidian, but I really don't like having conceallevel > 0

kmac:
@nobe4 I'm not sure if you've seen, but this repo was forked into https://github.com/obsidian-nvim/obsidian.nvim

I'm wondering if you might be up for re-submitting the PR there?  It's fairly active now.

I'm working around this by manually editing out the warning, but it's a pain to update via Lazy like this.

nobe4:
Good point! I'll open it over there

nobe4:
@kmac => https://github.com/obsidian-nvim/obsidian.nvim/pull/177

PR #802, opened 2024-12-31
Add `ObsidianSetCheckbox` (Attempt)

t-eckert:
Hi @epwalsh, I just started using your plugin today. It's fantastic!

I wanted to map an `ObsidianSetCheckbox x` to check off items for my day. Toggle is great, but it's not specifically what I want to use. 

I have the function check for the checkbox regex, grab the content of that checkbox and replace it with the value that is passed in.

### DIFF
--- a/lua/obsidian/util.lua
+++ b/lua/obsidian/util.lua
@@ -513,6 +513,7 @@ util.toggle_checkbox = function(opts, line_num)
   local checkboxes = opts or { " ", "x" }
 
   if not string.match(line, checkbox_pattern) then
+    -- Add an empty checkbox if one is not found.
     local unordered_list_pattern = "^(%s*)[-*+] (.*)"
     if string.match(line, unordered_list_pattern) then
       line = string.gsub(line, unordered_list_pattern, "%1- [ ] %2")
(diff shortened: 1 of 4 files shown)


lua/obsidian/util.lua
```diff
@@ -534,6 +535,22 @@ util.toggle_checkbox = function(opts, line_num)
...
+util.set_checkbox = function(check_char, line_num)
+  -- Allow line_num to be optional, defaulting to the current line if not provided
+  line_num = line_num or unpack(vim.api.nvim_win_get_cursor(0))
+  local line = vim.api.nvim_buf_get_lines(0, line_num - 1, line_num, false)[1]
+
+  local checkbox_pattern = "^%s*- %[.] "
+
+  if string.match(line, checkbox_pattern) then
+    -- Only set the check_char if the line contains the checkbox pattern.
+    line = util.string_replace(line, "- [" .. check_char .. "]", "- [" .. check_char .. "]", 1)
```
rodrigoscc: The issue seems to be here. You are trying to replace say `- [ x ]` with `- [ x ]`, so doing nothing. You need to replace `- [ <anything> ]` with `- [ <check_char> ]`.

rodrigoscc: You could extract `<anything>` with the regex `checkbox_pattern`, just make the char a regex group and extract it.

t-eckert: Thank you! I'll make this fix and test again. Great catch!

rodrigoscc (approved):
The new active fork is the following btw https://github.com/obsidian-nvim/obsidian.nvim. You may close this PR.

PR #803, opened 2025-01-02
fix: support pdf

jiyeol-lee:
now only url and pdf file types are supported to open but i think pdf file is pretty much common.


lua/obsidian/client.lua
LazyStability, line 1: The log.warn should read:
"This looks like a pdf path. You can customize the behavior of pdfs with the 'follow_pdf_func' option."

jiyeol-lee: Good catch, updated

PR #816, opened 2025-01-22
feat: add option for disabling frontmatter updates

thetwistedlogic:
While I do enjoy Obsidian creating yaml frontmatter for me, I've found strange issues with the fronmatter updating functionality. I've added the option to disable it.

### DIFF
--- a/lua/obsidian/client.lua
+++ b/lua/obsidian/client.lua
@@ -276,7 +276,7 @@ Client.templates_dir = function(self, workspace)
   return nil
 end
 
---- Determines whether a note's frontmatter is managed by obsidian.nvim.
+--- Determines whether a note's frontmatter is created by obsidian.nvim.
 ---
 ---@param note obsidian.Note
 ---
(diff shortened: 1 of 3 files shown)


lua/obsidian/client.lua
```diff
@@ -276,7 +276,7 @@ Client.templates_dir = function(self, workspace)
   return nil
 end
 
---- Determines whether a note's frontmatter is managed by obsidian.nvim.
+--- Determines whether a note's frontmatter is created by obsidian.nvim.
```
adamtajti, line 279: I think that this term should be kept as managed if both the update and the create functionality is going to be supported.

lua/obsidian/config.lua
```diff
@@ -18,7 +18,8 @@ local config = {}
 ---@field follow_url_func fun(url: string)|?
 ---@field follow_img_func fun(img: string)|?
 ---@field note_frontmatter_func (fun(note: obsidian.Note): table)|?
----@field disable_frontmatter (fun(fname: string?): boolean)|boolean|?
```
adamtajti, line 21: I think that this is a breaking change. The pull request [misses an update](https://github.com/epwalsh/obsidian.nvim/blob/main/.github/CONTRIBUTING.md#keeping-the-changelog-up-to-date) to the [`CHANGELOG.md`](https://github.com/epwalsh/obsidian.nvim/blob/main/CHANGELOG.md), but I assume that this would warrant a major version bump.

The other approach would be to prioritize `disable_frontmatter`. If it's set to true, or if it's a function that returns true, then that should make `disable_frontmatter_creation` and `disable_frontmatter_update` default to true as well, even if they are configured.

lua/obsidian/config.lua
```diff
@@ -52,6 +53,7 @@ config.ClientOpts.default = function()
     follow_url_func = nil,
     note_frontmatter_func = nil,
     disable_frontmatter = false,
+    update_frontmatter = false,
```
adamtajti, line 56: - The `ClientOpts` documentation update seems to signal that `disable_frontmatter` is no longer supported, yet it's set to false.
- The `ClientOpts` documentation doesn't seem to mention `update_frontmatter` at all, instead it has `disable_frontmatter_creation` and `disable_frontmatter_update`.
- The `README.md` file still references `disable_frontmatter` and it doesn't mentions the new settings.

adamtajti (commented):
I'm not a maintainer, but I skimmed through a couple of pull requests before planning to create one of my own and I left a couple of comments here that could be of use to improve this work.

thetwistedlogic:
> I'm not a maintainer, but I skimmed through a couple of pull requests before planning to create one of my own and I left a couple of comments here that could be of use to improve this work.

Oh wow. Thank you so much! I've actually moved away from this plugin completely. I prefer something that is a simple cli tool with neovim integration instead. Currently developing a homebaked tool.

This plugin unfortunately destroys the frontmatter for my other notes since it does not support yaml tags; this distrupts my workflow personally.

Regardless, thank you.

PR #817, opened 2025-01-27
feat(completion): blink.cmp support

### LINKED ISSUE #770: Support blink.nvim Autocomplete

Currently obsidian.nvim does not work with blink.nvim, only with nvim-cmp. However, as blink is gaining popularity and becoming the default for many, compatibility between blink and obsidian would be very welcome.

### Alternatives

_No response_

### Additional context

_No response_

adamtajti:
Adds support for the `blink.cmp` completion plugin. The `nvim_cmp` sources were refactored to reduce code duplication. The source/provider injection to the Blink configuration was rather hacky, but it works with functions as well. I tested the functionalities with both plugins, I was able to reference existing notes, with both syntax, create a new note, and add tags, although I might have missed something.

Closes https://github.com/epwalsh/obsidian.nvim/issues/770


adamtajti:
Closing this down as @guspix created a fork and duplicated this PR there:
https://github.com/obsidian-nvim/obsidian.nvim/pull/5

PR #837, opened 2025-03-07
FEAT: Allow users to pass blink source config up front

### LINKED ISSUE #20: Allow users to pass blink source config up front

The current implementation hard codes the passed source config table [here](https://github.com/obsidian-nvim/obsidian.nvim/blob/624dab41738881d749a4a75bf2e46ef98d4ca527/lua/obsidian/completion/plugin_initializers/blink.lua#L14C6-L27C4). I propose to allow users to pass this in their config as in below and then have it merged with the existing table

```
    opts = {
      completion = {
        nvim_cmp = false,
        blink = {
            enabled = true,
            score_offset = 10,
            transform_items = function(_, items)
                for _, item in ipairs(items) do
                    item.kind = 10
                end
             return items
          end,
        },
        min_chars = 0,
      },
    },
```

### Alternatives

For those who want this feature, it is possible to override the config after it has been registered with blink by adding a table in the providers table called obsidian and obsidian_tags which overrides these settings. The reason this would be better implemented in this plugin is it breaks the dependency between obsidian and blink on the config level and allows blink and obsidian to be individually lazy loaded.

### Additional context

Things to consider
    - what if a user wants to customize obsidian and obsidian_tags separately?
    - what would happen to the nvim-cmp config?
    - is this worth it when you could just couple blink and obsidian at the plugin declaration level?

GersiD:
Closes https://github.com/obsidian-nvim/obsidian.nvim/issues/20#issue-2901858088


GersiD:
Wrong repo

PR #854, opened 2025-05-13
Cmp completion filename 

RokPre:
If a note has an id, it does not show up in the cmp dropdown menu, if you type in "[[<filename>]]".

This adds it.


RokPre:
:)

PR #855, opened 2025-05-18
Handle non-suffix decimals in note ID

bburgess19:
A reasonable use case is the

[Johnny.Decimal](https://johnnydecimal.com/) system.

In this case, one might have a folder with Zettelkasten notes in:

`/30 Zettelkasten/31 Atomic/31.01 - 1747366626-my-insight`

but the previous implementation would, when creating a new note, make the path

`/30 Zettelkasten/31 Atomic/31.md`

because it considered everything after the "31" to be the suffix.

This PR fixes this.


neo451:
this repo is not maintained at the moment, check out https://github.com/obsidian-nvim/obsidian.nvim, we would be happy to work on things like this :)

bburgess19:
@neo451  You're the goat!

PR #870, opened 2026-02-14
Removed tracker from YouTube link

zperk13:
A while ago, YouTube made it so when you copy a YouTube video link using the share button, it adds a tracker at the end of it (the `?si=8nbnNsRVnw0xfX2S` thing). It is not necessary at all for the link to work, and only exists for Google to collect more data it doesn't need about us. This PR just removes the tracker.


neo451:
@zperk13 hi, this repo is not maintained anymore, checkout: https://github.com/obsidian-nvim/obsidian.nvim

